PR #233, opened 2015-02-26
Fixes #232 - ArgumentNullException thrown by Identify(string userId)

### LINKED ISSUE #232: [iOS] ArgumentNullException: Argument cannot be null. Parameter name: userId
Raygun is initialized like this:
```c#
static void Main(string[] args) {
    RaygunClient.Attach("API_KEY", true, true);
            
    UIApplication.Main(args, null, "AppDelegate");
}
```

Stack trace:

    [ArgumentNullException: Argument cannot be null. Parameter name: userId]
    Mindscape.Raygun4Net.Xamarin.iOS.Raygun.Identify(System.String userId) in <filename unknown>:0
    Mindscape.Raygun4Net.RaygunClient.Attach(System.String apiKey, System.String user, Boolean canReportNativeErrors, Boolean hijackNativeSignals) in <filename unknown>:0
    Mindscape.Raygun4Net.RaygunClient.Attach(System.String apiKey, Boolean canReportNativeErrors, Boolean hijackNativeSignals) in <filename unknown>:0

Likely cause: https://github.com/MindscapeHQ/raygun4net/blob/v4.2.0/Mindscape.Raygun4Net.Xamarin.iOS/RaygunClient.cs#L249

jamie94bc:
Fix based on existing check for `DeviceId` in `BuildRaygunIdentifierMessage()`.

https://github.com/MindscapeHQ/raygun4net/blob/v4.2.0/Mindscape.Raygun4Net.Xamarin.iOS/RaygunClient.cs#L474

### DIFF
--- a/Mindscape.Raygun4Net.Xamarin.iOS/RaygunClient.cs
+++ b/Mindscape.Raygun4Net.Xamarin.iOS/RaygunClient.cs
@@ -353,9 +353,11 @@ namespace Mindscape.Raygun4Net
       }
 
       _client.User = user; // Set this last so that it can be passed to the native reporter.
-      if (user == null && _client._reporter != null)
+	  
+      string deviceId = _client.DeviceId;	  
+      if (user == null && _client._reporter != null && !String.IsNullOrWhiteSpace(deviceId))
       {
-        _client._reporter.Identify(_client.DeviceId);
+        _client._reporter.Identify(deviceId);
       }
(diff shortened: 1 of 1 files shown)


QuantumNightmare:
Thanks for the PR. I've merged this in so it will be included in the next version - planned for next week. I'll also look into why the DeviceId is null.

PR #261, opened 2015-07-16
Use InnerException not GetBaseException

WaldenL:
GetBaseException can send you too far down the rabbit hole. For example in a SQL Timeout exception the BaseException is a useless "[Win32Exception: The wait operation timed out]" without a stack trace as it was captured and wrapped by SqlClient. All we want to do here is remove the outermost HttpUnhandledException wrapper

### DIFF
--- a/Mindscape.Raygun4Net/RaygunHttpModule.cs
+++ b/Mindscape.Raygun4Net/RaygunHttpModule.cs
@@ -78,7 +78,7 @@ namespace Mindscape.Raygun4Net
     {
       if (exception is HttpUnhandledException)
       {
-        return exception.GetBaseException();
+        return exception.InnerException;
       }
 
       return exception;


QuantumNightmare:
Thanks for the PR. We'll discuss this and get back to you.

QuantumNightmare:
Thanks again for the PR. When changing GetBaseException to InnerException, the Unwrap method becomes somewhat redundant - as we have a RaygunClient level feature where you can specify wrapper exception to be removed. One of the defaults is in fact HttpUnhandledException.

As such, I've created a new PR that removes Unwrap entirely which will behave the way you'd expect. We've released a new version of Raygun4Net (v5.2.0) which includes this. Here's the PR if you want to see: https://github.com/MindscapeHQ/raygun4net/pull/273/files

Please let me know if you have further questions about this.

PR #263, opened 2015-08-29
Changes Client name to Raygun4Net

nelsonsar:
In order to make Raygun work for Xamarin.iOS the client message is now using ```Raygun4Net``` as client name, since the last one, ```Raygun4Net.Xamarin.iOS.Unified```, was not been reported in the dashboard.

P.S.: I nerver worked with C# before, so sorry if I broke something in your ```csproj``` file :/


QuantumNightmare:
Thanks for the PR. There was an issue at the other end for receiving Xamarin.iOS (classic and unified) reports which has now been solved. I'm sorry to say, that means we will not need to merge this PR. The client name is used to select an appropriate exception grouping logic, which Xamarin.iOS has a couple of special cases.

Great work finding this issue. This PR helped me to track it down at our end. Apologies for the trouble.

PR #269, opened 2015-10-25
Made RaygunHttpModule extendable

jcvandan:
I have been using Raygun for a while, and I liked the way I can just use add the nuget package and have all my MVC errors logged in a centralized location without me having to worry about all the potential points of failure in an application. However, I realised I wanted to add some common custom data to all my error logs, i.e. username and session id. I didn't want to start implementing custom logging all over the place so my first port of call was to try and implement a custom HttpModule for the logging, but when I tried this I saw that the module is not extendable.

To solve this I have simply made the `SendError` methods protected virtual, and have made them call some protected methods `SendErrorInBackground` (these methods contain the original logging logic). The `SendErrorInBackground` methods now take a customData Dictionary parameter, which means it would be very east to implement a custom http module which overrode the `SendError` methods and passed whatever custom data was needed to the new `SendErrorInBackground` and everything would work the same as previously.

I realise there may be reasons for not making this class extendable, but as it is such a small change I thought it would be worth doing it and doing a pull request. If the pull request is not accepted I will just created my own nuget package and run my fork alongside the master branch and periodically merge (although I would prefer not to do this!!).

### DIFF
--- a/Mindscape.Raygun4Net/RaygunHttpModule.cs
+++ b/Mindscape.Raygun4Net/RaygunHttpModule.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Web;
@@ -32,7 +33,17 @@ namespace Mindscape.Raygun4Net
     {
     }
 
-    private void SendError(object sender, EventArgs e)
+    protected virtual void SendError(object sender, EventArgs e)
+    {
+      SendErrorInBackground(sender, e);
+    }
+
+    protected virtual void SendError(HttpApplication application, Exception exception)
+    {
+      SendErrorInBackground(application, exception);
+    }
+
+    protected void SendErrorInBackground(object sender, EventArgs e, Dictionary<string, object> customData = null)
     {
       var application = (HttpApplication)sender;
       var lastError = application.Server.GetLastError();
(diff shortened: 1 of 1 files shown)


QuantumNightmare:
Hi jcvandan,

Apologies for the late reply. Before merging this, could you please have a try at using the SendingMessage event to add the custom data? To do this, first you'd provide your own RaygunClient instance to the Http module, which is explained here: https://raygun.io/docs/languages/net/asp#raygun-client

Then, on your RaygunClient instance, you can listen to the SendingMessage event. The event handler will be called for every error report that gets sent to Raygun. Here you can add custom data to e.Message.Details.UserCustomData (make sure to check for null when going through this structure). There is an example of using the SendingMessage event here: https://raygun.io/docs/languages/net/asp#modify-cancel-message (The example is using the event to cancel a message).

Please try that out and let me know how it goes for you. We are also considering adding an option to Raygun4Net to set tags and custom data that gets sent with every message. Please let me know if you have any questions.

jcvandan:
@QuantumNightmare thanks for the reply, I will have a look at doing this in the next few days when I get a spare moment. I will report back!

mduncan26:
Closing this PR for now. If you find this issue to still be outstanding and wish to see this PR addressed, please create a new one or post within our forums (https://raygun.com/forums). Thank you.

PR #295, opened 2016-06-04
Flag as sent after send. Resolves #294

### LINKED ISSUE #294: SendInBackground with Exception.Data sends AlreadySentByRagun
``` csharp
var exception = new Exception()
ex.Data["Custom Data"] = "Test";

var client = new RaygunClient();
client.SendInBackground(exception, null, exception.Data);
```

The above code causes a piece of custom data `AlreadySentByRagun` to appear in Raygun.

nigel-sampson:
Flag the exception has sent after it's actually sent.

### DIFF
--- a/Mindscape.Raygun4Net/RaygunClient.cs
+++ b/Mindscape.Raygun4Net/RaygunClient.cs
@@ -304,6 +304,7 @@ namespace Mindscape.Raygun4Net
           {
             _currentRequestMessage = currentRequestMessage;
             Send(BuildMessage(exception, tags, userCustomData, userInfo, currentTime));
+            FlagAsSent(exception);
           }
           catch (Exception)
           {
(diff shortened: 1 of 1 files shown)


QuantumNightmare:
Hi Nigel,

Thanks for the PR, but it's best that the flag gets set outside the thread pool queuing. By setting the flag only after it's actually been sent, it is quite easy for an exception to be sent multiple times through the SendInBackground method - they can all be queued without the flag having been sent yet.

To solve the issue with this flag showing up in your tags list, the best solution would be for the Raygun4Net provider to not send the flag. I'm not sure when we'll implement this, but feel free to update this PR. Another option immediatley available to you would be to listen to when Raygun is about to send a message, and then strip this flag out. Here's the docs to do so: https://raygun.com/docs/languages/net/asp#modify-cancel-message

mduncan26:
Closing this PR for now. If you find this issue to still be outstanding and wish to see this PR addressed please create a new one or post within our forums (https://raygun.com/forums). Thank you.

PR #300, opened 2016-06-19
Use ClickOnce data directory as base if application is network deployed

anderssonjohan:
This solves the problem having two client applications running on the same machine using Raygun.
When the network dies all messages for all applications using Raygun stores their reports in the
same directory by default. The reports are sent to Raygun at a later point in time, by the first
application to start up with the Raygun client.

This leads to exception reports being sent to the wrong Raygun account and by isolating the
reports stored on the local hard drive we can at least get the right reports into the right
account at Raygun.io.

### DIFF
--- a/Mindscape.Raygun4Net4/RaygunClient.cs
+++ b/Mindscape.Raygun4Net4/RaygunClient.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


QuantumNightmare:
Thanks again for your help with this issue. For reference, here is the PR that this resulted in: https://github.com/MindscapeHQ/raygun4net/pull/301

Closing this PR now.

PR #314, opened 2016-12-09
Only read text requests

### LINKED ISSUE #313: RaygunWebApiDelegatingHandler can lead to out of memory errors
The handler reads every request into memory as a string, so that it may record the contents. This is bad for very large requests, like file uploads. We use the WebApi to accept file uploads, and have encountered OutOfMemoryExceptions due to Raygun. I would suggest only reading the request if the Content-Type is known to be json, xml, or text.

pjhuck:
This fixes #313 -- which prevents Raygun from reading large binary requests (like file uploads) into memory.


QuantumNightmare:
Hi pjhuck, Thanks for submitting this PR, we'll look at merging this in for the next version of Raygun4Net.

UberMouse:
I've gone with a slightly different approach to solving this rather than just whitelisting some specific content types. Instead of reading the entire Body into a string and then trimming it I'm reading it as a stream and only reading at most the first 4096 bytes which should be a bit safer overall. Also can't run into any potential issues about missing some whitelisted content types. See #329

PR #317, opened 2016-12-29
Allowing Raw Data capture for ASP.NET Core

bgardner87:
I don't think this is necessarily ready for merge but I wanted to get it shared.....

If my understanding is correct in how ASP.NET Core handles requests, then the raw data handling currently in raygun is broken WRT ASP.NET Core in the following ways:

1. Non-Form streams are not rewindable, so this line:
request.Body.Seek(0, SeekOrigin.Begin);
breaks as Seek is not supported, and the stream has already been read. To fix, I implemented an update to the middleware that replaces the incoming stream with one that has CanSeek == true. This is controlled by a new property "TransformRequestStream"

2. After we have a rewindable stream, the code that strips out ignored properties uses request.Form, which is not populated for non urlencoded form posts. It looks like the code to do this is dead code in ASP.NET core, as it is only called when the content type is not a form, so request.Form will always be empty (i think).

For my fix, I removed the function calls that strip the ignore properties. This means that the feature would be completely disabled for ASP.NET Core, though I don't believe it was working at all to begin with based on the above.

Next steps would be to re-implement the ignored fields stripping, but need to take into account the content type, as request.Form is no longer reliable across content types. This would likely include deserializing based on type and working from there.

### DIFF
--- a/Mindscape.Raygun4Net.AspNetCore/Builders/RaygunAspNetCoreRequestMessageBuilder.cs
+++ b/Mindscape.Raygun4Net.AspNetCore/Builders/RaygunAspNetCoreRequestMessageBuilder.cs
@@ -96,9 +96,15 @@ namespace Mindscape.Raygun4Net.Builders
           request.Body.Seek(0, SeekOrigin.Begin);
           string temp = new StreamReader(request.Body).ReadToEnd();
 
+          /*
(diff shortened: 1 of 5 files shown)


QuantumNightmare:
Thanks for the PR. This functionality has been merged in and is available right now in version 5.5. You'll need to set the replaceUnseekableRequestStreams config setting to true for this to take effect.

PR #320, opened 2017-02-14
Fix ignore JavaProxyThrowable handling in recent Xamarin releases

jamie94bc:
In recent versions of Xamarin (not sure on exact release, but I assume Xamarin.Android 6.1 as the JNI architecture was changed then) the first line in the stack trace had changed to include the top level throwable class name rather than "exception".

Added handling for proxy stack traces in Xamarin.Android 6.1(?)+ and associated tests.

Also fixed handling for non-wrapped throwables which will have `JavaProxyThrowable` on line 2 of the stack trace, and no `Caused by: ` prefix.

##### Wrapped #####
```
    --- End of managed Java.Lang.RuntimeException stack trace ---
 java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
 com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120):0
 Caused by: java.lang.reflect.InvocationTargetException
 java.lang.reflect.Method.invoke(Native Method):0
 com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230):0
    ... 1 more
 Caused by: android.runtime.JavaProxyThrowable: System.NullReferenceException: Object reference not set to an instance of an object
```
##### Non-wrapped #####
```
    --- End of managed Android.Runtime.JavaProxyThrowable stack trace ---  
 android.runtime.JavaProxyThrowable: System.NullReferenceException: Object reference not set to an instance of an object
```

### DIFF
--- a/Mindscape.Raygun4Net.Xamarin.Android.Tests/RaygunClientTests.cs
+++ b/Mindscape.Raygun4Net.Xamarin.Android.Tests/RaygunClientTests.cs
@@ -2,10 +2,11 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Text;
-using System.Threading.Tasks;
 using Mindscape.Raygun4Net.Messages;
 using NUnit.Framework;
+using Java.Lang;
+using Java.Lang.Reflect;
+using Exception = System.Exception;
 
 namespace Mindscape.Raygun4Net.Xamarin.Android.Tests
 {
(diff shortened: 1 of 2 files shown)


Filip-Dimitrievski:
This pull request looks great! I will test this later today and will look to integrate it in after that.

I will get back to you with an update on how that goes. 

We really appreciate pull requests like this. Thanks for looking to improve Raygun4Net!

QuantumNightmare:
Thanks for the PR @jamie94bc, I'm merging this in now to be included in the next version of Raygun4Net.

PR #324, opened 2017-03-15
Raygun4net webjobs sdk

UberMouse:
Adds first class support for Azure WebJobs run using Azure Web Jobs SDK, which require custom exception handling compared to running a non Web Jobs SDK enabled console application for example.

---
This change is [Reviewable](https://reviewable.io/reviews/mindscapehq/raygun4net/324)

### DIFF
--- a/Mindscape.Raygun4Net.Azure.WebJob/JobHostConfigurationExtensions.cs
+++ b/Mindscape.Raygun4Net.Azure.WebJob/JobHostConfigurationExtensions.cs
@@ -19,5 +19,10 @@ namespace Mindscape.Raygun4Net.Azure.WebJob
 
       config.Tracing.Tracers.Add(traceMonitor);
     }
+
+    public static void UseRaygun(this JobHostConfiguration config)
+    {
+      UseRaygun(config, RaygunSettings.Settings.ApiKey);
+    }
   }
 }
--- a/Mindscape.Raygun4Net.Azure.WebJob/JobHostConfigurationExtensions.cs
+++ b/Mindscape.Raygun4Net.Azure.WebJob/JobHostConfigurationExtensions.cs
@@ -10,12 +10,6 @@ namespace Mindscape.Raygun4Net.Azure.WebJob
 {
   public static class JobHostConfigurationExtensions
   {
-    public static void UseRaygun(this JobHostConfiguration config, string apiKey)
-    {
-      var client = new RaygunClient(apiKey);
-      UseRaygun(config, client);
-    }
-
     public static void UseRaygun(this JobHostConfiguration config, RaygunClient client)
     {
       var processor = new RaygunExceptionHandler(client);
@@ -24,6 +18,12 @@ namespace Mindscape.Raygun4Net.Azure.WebJob
         .Subscribe(processor.Process);
 
       config.Tracing.Tracers.Add(traceMonitor);
+    }
+
+    public static void UseRaygun(this JobHostConfiguration config, string apiKey)
+    {
+      var client = new RaygunClient(apiKey);
+      UseRaygun(config, client);
     }
 
     public static void UseRaygun(this JobHostConfiguration config)
--- a/Mindscape.Raygun4Net.Azure.WebJob/JobHostConfigurationExtensions.cs
+++ b/Mindscape.Raygun4Net.Azure.WebJob/JobHostConfigurationExtensions.cs
@@ -1,8 +1,3 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Azure.WebJobs.Extensions;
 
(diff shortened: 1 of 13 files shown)


xt0rted:
Adding Raygun to my webjobs has been on my backlog for way too long. If there's a pre-release build of this I can try from MyGet or something I'll gladly try it out.

UberMouse:
There should be something for you to try out tomorrow NZ time (19:40 currently). I've never actually used WebJobs before, does this approach look like it makes sense? I'm basically doing something similar to what I saw the WebJob.Extensions NuGet package do, by hooking in via the JobHostConfiguration object. Like so

```C#
var config = new JobHostConfiguration();
config.UseRaygun()

var host = new JobHost(config);
host.RunAndBlock()
```

And it will report any exceptions thrown inside your WebJob code back to Raygun. You can optionally provide your own RaygunClient instance if you wish to attach any handlers/do extra configuration etc.

Are there any use cases you might want that aren't covered by this?

xt0rted:
I was looking into using the [ErrorTrigger]( https://github.com/Azure/azure-webjobs-sdk-extensions#errortrigger) and adding one per webjob project but I like the `JobHostConfiguration` route much better. It's much easier to implement and far less code to remember.

There are a couple [environment variables](https://github.com/projectkudu/kudu/wiki/WebJobs#environment-settings) that you can use to get the webjob's name & run ID but I'm not sure if there's any real difference between those and the way you've implemented it. I run these a lot outside of azure during development and testing so trying to pull this from the `TraceFilter` is probably better.

If an exception is thrown before the function runs it's possible for nothing to be sent. To handle exceptions on startup you'd need to add a handler onto `AppDomain.CurrentDomain.UnhandledException` or wrap the startup code in a try/catch.

Mindscape.Raygun4Net.Azure.WebJob/JobHostConfigurationExtensions.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
```
xt0rted: All of the `System` usings are unused

Mindscape.Raygun4Net.Azure.WebJob/Mindscape.Raygun4Net.Azure.WebJob.csproj
```diff
@@ -0,0 +1,134 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\Mindscape.Raygun4Net\Mindscape.Raygun4Net.csproj">
+      <Project>{495e53b3-f3af-4c4f-baaf-865efaa2f4a9}</Project>
+      <Name>Mindscape.Raygun4Net</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="app.config" />
+    <None Include="packages.config" />
+    <None Include="Raygun4Net.Azure.WebJob.snk" />
```
xt0rted: This file isn't included in the project

Mindscape.Raygun4Net.Azure.WebJob/RaygunProcessor.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
```
xt0rted: This and the next using aren't used

Mindscape.Raygun4Net.Azure.WebJob/RaygunProcessor.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <param name="filter">The <see cref="TraceFilter"/> containing information about the exception.</param>
+    public void Process(TraceFilter filter)
+    {
+      // Get Web Job scm Host URI: yoursite.scm.azurewebsites.net
+      var httpHost = Environment.GetEnvironmentVariable("HTTP_HOST");
+
+      var events = filter.GetEvents().Where(e => e.Exception != null);
+      foreach (var traceEvent in events)
+      {
+        var tags = new List<string>();
```
xt0rted: Similar to the JS client I've started adding `UnhandledException` to my unhandled exception tags. Is this something that would be valuable to add in here as well?

xt0rted:
Not sure if you've seen it but in https://github.com/Azure/azure-webjobs-sdk/pull/1059 they're adding first class logging to WebJobs & Functions. One of the default loggers is Application Insights which also does exception logging. Once that's merged Raygun should plug right in. They're also using the `MS_Function` keys so my previous comment on the environment variables can be ignored.

Mindscape.Raygun4Net.Azure.WebJob/RaygunProcessor.cs
UberMouse: Hmm, I see the merit in the idea but it's inconsistent with how the rest of the providers do it so I'd prefer to keep consistency with them. I have added support for setting a `CustomTags` array directly on the RaygunExceptionHandler if you wish to add that yourself however.

UberMouse: Actually was wrong here, adding UnhandledException is in line with the other providers. I meant to do it before the release but I forgot so I'll try and roll it into the next release. You can add it your self for now at least.

UberMouse:
The NuGet for this is released onto NuGet.org https://www.nuget.org/packages/Mindscape.Raygun4Net.Azure.WebJob/5.4.1

Issue #322, opened 2017-02-17
Request for ASP.NET Core: allow HttpClient configuration


FLGMwt:
Before ASP.NET Core, we configured connection limits and SSL protocols using the static `ServicePointManager`:

```
ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls12 | SecurityProtocolType.Tls11;
ServicePointManager.DefaultConnectionLimit = 60;
```

`ServicePointManager` is no longer available and it looks like the [only way](http://stackoverflow.com/questions/39560249/servicepointmanager-in-asp-net-core) to configure this is at the instantiation of the `HttpClient`. 

I'm not *quite* sure what the configuration would look like, but given I hope #321 also gets in, it would have to be at initial setup.

Obi-Dann:
cc @phillip-haydon there's a simple PR available fixing this issue #507

Issue #321, opened 2017-02-17
`HttpClient`s should not be new'd up on every send


FLGMwt:
It looks like all the `RaygunClient.Send`s new up an `HttpClient` for each request. According to [Microsoft](https://github.com/mspnp/performance-optimization/blob/master/ImproperInstantiation/docs/ImproperInstantiation.md), [this guy](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/), and our experience doing the same thing with `HttpClient`s at scale, this is a bad thing because of socket exhaustion.

Ideally, it would be nice to have a hook for the consuming application to provide a singleton `HttpClient` instance, but at the very least, the `HttpClient` should be a singleton within the raygun4net library.

Happy to jump in on this : )

CmdrKeen:
Thanks for this, we'll incorporate it into our next update. Now that .NET Core has been updated along with the VS 2017 release, we'll issue an update soon. 

Appreciate your good spotting on this @FLGMwt !

DanFTRX:
Has any progress been made on this?  This seems like a serious attack vector on any web server running Raygun as if you can induce raygun errors, it can starve the socket supply.

DanFTRX:
Is this project still under active development?

QuantumNightmare:
Yes, we still work on this project when we can. We record each issue as tickets in our back log along with tasks for all other parts of Raygun. These then get prioritized to specify the order that we work on them. We are not up to this task yet.

grugcrood82:
Resolved by #409

PR #327, opened 2017-03-16
Singleton http client

UberMouse:
Instead of creating a new HttpClient every time an exception is sent, share a singleton instance as per #321. Make it a static Property with a setter so it can be supplied by the user as per #322.

---
This change is [Reviewable](https://reviewable.io/reviews/mindscapehq/raygun4net/327)

### DIFF
--- a/Mindscape.Raygun4Net.WebApi/packages.config
+++ b/Mindscape.Raygun4Net.WebApi/packages.config
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.AspNet.WebApi.Client" version="5.1.1" targetFramework="net45" />
-  <package id="Microsoft.AspNet.WebApi.Core" version="5.1.1" targetFramework="net45" />
-  <package id="Newtonsoft.Json" version="4.5.11" targetFramework="net45" />
+  <package id="Microsoft.AspNet.WebApi.Client" version="5.2.3" targetFramework="net45" />
+  <package id="Microsoft.AspNet.WebApi.Core" version="5.2.3" targetFramework="net45" />
+  <package id="Newtonsoft.Json" version="6.0.4" targetFramework="net45" />
 </packages>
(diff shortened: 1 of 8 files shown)


Mindscape.Raygun4Net.WebApi/packages.config
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.AspNet.WebApi.Client" version="5.1.1" targetFramework="net45" />
-  <package id="Microsoft.AspNet.WebApi.Core" version="5.1.1" targetFramework="net45" />
-  <package id="Newtonsoft.Json" version="4.5.11" targetFramework="net45" />
+  <package id="Microsoft.AspNet.WebApi.Client" version="5.2.3" targetFramework="net45" />
+  <package id="Microsoft.AspNet.WebApi.Core" version="5.2.3" targetFramework="net45" />
+  <package id="Newtonsoft.Json" version="6.0.4" targetFramework="net45" />
```
j5alive: Did these packages need to be updated? Do we even use Newtonsoft.Json in this project?

Mindscape.Raygun4Net.WinRT/RaygunClient.cs
```diff
@@ -336,9 +341,9 @@ protected IEnumerable<Exception> StripWrapperExceptions(Exception exception)
     private async Task PostMessageAsync(HttpClient client, HttpContent httpContent, Uri uri)
 #pragma warning restore 1998
     {
+      // Unsure why this is structured like this, don't want to change incase it's for a valid reason
```
j5alive: I would say the statements can be joined and then you can remove this comment.

Mindscape.Raygun4Net.WebApi/packages.config
UberMouse: Had to update WebApi.Core to work with the new WebApi version (Raygun4Net.WebApi didn't work in my brand new WebApi project) and I believe that WebApi.Client depends on NewtonSoft.Json >= 6.0.4

grugcrood82:
Resolved by #409

BarsikV:
This was first resolved by #409 , but then the PR was reverted by #410 . Please fix the issue with HttpClient and make it Singleton, static or configurable. It's harming our services.

PR #353, opened 2017-08-29
ASP.NET Core: prevent nulling out existing HttpContext  (fixes #352)

### LINKED ISSUE #352: ASP.NET Core: calls to Send can null out current HttpContext
It looks like there are some funny flows you can end up going down when using the ASP.NET Core version of `RaygunClient`. In the one I ran into, I was creating a new client by [passing in the context to the constructor](https://github.com/MindscapeHQ/raygun4net/blob/master/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs#L86). All of the `Send` overloads that _don't_ take a `HttpContext` parameter result in eventually passing a `null` value for `HttpContext` into `SendAsync`, which proceeds to [wipe out the previous value stored in the client](https://github.com/MindscapeHQ/raygun4net/blob/master/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs#L338).

Is this intentional, or should it do a `null` check before assigning that value? I'd be happy to throw up a PR for it too, but wanted to check if I was missing something first. 

In the `SendInBackground` flow it gets [read out first in `BuildRequestMessage`](https://github.com/MindscapeHQ/raygun4net/blob/master/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs#L409) and added to the message before being nulled out.

gshackles:
This addresses issue #352. Let me know if I missed something here!

### DIFF
--- a/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
+++ b/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
@@ -335,7 +335,11 @@ namespace Mindscape.Raygun4Net
     {
       if (CanSend(exception))
       {
-        _currentHttpContext.Value = context;
+        if (context != null)
+        {
+          _currentHttpContext.Value = context;
+        }
+
         _currentRequestMessage.Value = await BuildRequestMessage();
         _currentResponseMessage.Value = BuildResponseMessage();
 


davidrevoledo:
any news of this ? :) I really need, I don't know if I can help to do something but it'd be really good to have the http context information in asp.net core webapps. thanks

PR #358, opened 2017-09-09
Don't read ignored response content.

kogir:
If it's not going to be sent, there's no reason to read it at all.


QuantumNightmare:
Hi kogir,

Thanks the PR, I've merged this in to be included in the next release.

-Jason Fauchelle

Issue #347, opened 2017-07-20
RaygunClient.SendInBackground() does not set Version to AssemblyVersion in ASP.NET


essizershinn:
It appears that `RaygunClient.SendInBackground()` cannot get the AssemblyVersion in ASP.NET because it calls [RaygunClient.BuildMessage()](https://github.com/MindscapeHQ/raygun4net/blob/master/Mindscape.Raygun4Net/RaygunClient.cs#L369) from a [ThreadPool queued work item](https://github.com/MindscapeHQ/raygun4net/blob/master/Mindscape.Raygun4Net/RaygunClient.cs#L308) so `HttpContext.Current` is null when `RaygunMessageBuilder.SetVersion(string)` looks at it.

Repro Steps
1. Create an instance of `RaygunClient` without an explicit ApplicationVersion (either in config or set in code).
2. Call `RaygunClient.SendInBackground(...)` from within an ASP.NET application.

Expected Behavior
- Version is set to AssemblyVersion

Actual Behavior
- Version is set to **Not Supplied**

Configuration
- Raygun4Net4 v5.2
- .NET 4.6.1 Full Framework / WebApi 2.2 or MVC 5
- Windows 10 Creators Update (1703) / IIS 10.0.15063.0

QuantumNightmare:
Hi essizershinn

Thanks for reporting this issue. We'll take a look at resolving this for the next version of Raygun4Net.

phillip-haydon:
This should no longer be an issue, if no app version is provided it will roll back to assembly version.

Issue #297, opened 2016-06-07
Events on RaygunClient before entering background thread


gregpakes:
Hi,

I have a requirement where I want to intercept an event on RaygunClient.  I have looked at CustomGroupingKey and SendingMessage events, but when used with the HttpModule, these fire on a background thread, after SendInBackground is called.

I have forked and have a working version with a new event called BeforeSendingMessage which makes sure the event is fired before calling the background thread.

Is this something that you would be interested in me creating a PR for?

If so, do I need to go and add all the functionality in all the RaygunClient.cs files for each project?

thanks,

Greg

QuantumNightmare:
Hi Greg,

Apologies for the late reply. Feel free to create a PR, we can take a look at it when working on the next version. Don't worry about adding this support to all the .NET providers, we could do that.

-Jason Fauchelle

mduncan26:
Closing this issue for now. If you wish to see this issue addressed please create a new issue or post within our forums (https://raygun.com/forums). Thank you.

PR #361, opened 2017-10-12
RaygunMessage Initializer Support

Bio2hazard:
Hey!

We use raygun a lot and love it, but ran into a lot of restrictions when trying to enrich our exception reports with information residing on the current thread.

Other libraries we use commonly have the concept of Initializers - customer-added code that runs directly after the message is created.

I have recreated that functionality by building out support for it in the base Raygun4Net project, and then implemented it in the Raygun4Net.WebApi project. 

In order to ensure the initializers run on correct thread, I had to move the Threadpool to send the message to Raygun down the stack, after the message has been built. This also allows for removing the ThreadLocal RaygunRequestMessage, as we can now build it directly in the BuildMessage call.

I matched your code standards to the best of my abilities and documented all public calls. I also added a unit test that shows how to use it, and verifies the behavior.

Please let me know if you have any questions, or need me to change something!

This would also solve #347 & #297

### DIFF
--- a/Mindscape.Raygun4Net.Core/IRaygunMessageInitializer.cs
+++ b/Mindscape.Raygun4Net.Core/IRaygunMessageInitializer.cs
@@ -1,0 +1,20 @@
+﻿using Mindscape.Raygun4Net.Messages;
+
+namespace Mindscape.Raygun4Net
+{
+  /// <summary>
+  /// Represents an initializer for <see cref="RaygunMessage"/> objects.
+  /// </summary>
+  /// <remarks>
+  /// The Raygun Clients use registered <see cref="IRaygunMessageInitializer"/>
+  /// to automatically add additional properties to <see cref="RaygunMessage"/> objects.
+  /// </remarks>
+  public interface IRaygunMessageInitializer
+  {
+    /// <summary>
+    /// Initializes properties of the specified <see cref="RaygunMessage"/> object.
+    /// </summary>
+    /// <param name="raygunMessage">The <see cref="RaygunMessage"/> to initialize.</param>
+    void Initialize(RaygunMessage raygunMessage);
+  }
+}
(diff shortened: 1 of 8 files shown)


QuantumNightmare:
Hi Bio2hazard,

Thanks for the PR, we'll take a look when we have time and hopefully include it in the next release.

I've only had a quick look though this, it seems similar to the SendingMessage event that lets you perform additional logic when a message is about to be serialized and sent to Raygun. If you have not seen this, try it out and see if it can achieve the same behavior or not: https://raygun.com/docs/languages/net/webapi#modify-cancel-message

-Jason Fauchelle

Bio2hazard:
Hello Jason,

I am well aware of the SendingMessage event, but unfortunately it is not sufficient to meet my requirements ( and that of others, judging by the issue tracker ).

The key differences are:
- For SendInBackground requests, the SendingMessage event runs on the background thread, and as such does not have access to call context. This is a problem as it makes it impossible to enrich the exception with request-related information.
- Being an event handler, SendingMessage is limited in it's capabilities. Database access, dependency injection and calls to non-static application code is very difficult to safely accomplish.
- Adding the proposed Initializer is a non-breaking change allowing for an additional extensibility point using user-provided code. This gives infinite freedom to enrich the exception with information. It can be kept simple as a initializer that simply runs on the same thread and has access to the HttpContext, or complicated by registering a per-request scoped factory delegate on the applications' dependency injection container and using that for initialization. It gives the user the freedom to do what they need to get information that will aid them in debugging.

Right now I can't even re-use a single RaygunClient instance and set the affected user from SendingMessage because I don't have access to the context that contains the user information. I am forced to create a new raygun client every time and set the user, but that does not let me set Custom data or tags.

Thank you for your time. I am hopeful to see this in the next release as it would make embedding debugging information in the exception indefinitely easier.

phillip-haydon:
Closing this one, modern .NET is extensible via `IMessageBuilder` and classic Web captures the request before dispatching. The full extensibility probably wont be back ported to .NET Framework.

PR #377, opened 2018-04-20
XRAY-1359 Raygun4Net .NetCore2/.NetStandard2 Support

mduncan26:
Adding a new Raygun4Net provider to support .NetCore2/.NetStandard2


Mindscape.Raygun4Net.NetCore/Messages/RaygunClientMessage.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Mindscape.Raygun4Net
+{
+  public class RaygunClientMessage
+  {
+    public RaygunClientMessage()
+    {
+      Name = "Raygun4Net.NetCore";
```
jscottnz: should this be NetCore2?

mduncan26:
I have tested using the NuGet package for this provider in both .NetCore1 and .NetCore2 console applications without problems.

Mindscape.Raygun4Net.NetCore/Messages/RaygunClientMessage.cs
```diff
@@ -0,0 +1,18 @@
...
+      Version = "5.5.4";
```
QuantumNightmare: If possible, it would be good to work this out in code from the assembly version info like I think other Raygun4Net providers do. Great to see you added this location to the wiki though.

PR #380, opened 2018-05-03
XRAY-1450: Raygun4Net .NET Core 2 provider breadcrumb support.

mduncan26:
In memory breadcrumb support.


Mindscape.Raygun4Net.NetCore/Breadcrumbs/RaygunBreadcrumb.cs
```diff
@@ -0,0 +1,28 @@
...
+    
+    public RaygunBreadcrumbLevel Level { get; set; } = RaygunBreadcrumbLevel.Info;
+
+    public RaygunBreadcrumbType Type { get; set; } = RaygunBreadcrumbType.Manual;
+    
+    public string Message { get; set; }
+
+    public string Category { get; set; }
+
+    public IDictionary<string, object> CustomData { get; set; } = new Dictionary<string, object>();
```
snakefoot: Should not allocate Dictionary-object upfront. It should be provided from the outside.

Issue #288, opened 2016-02-11
Request body not logged for owin


Fnopster:
Hi,

I have a web api project that uses owin. In the startup, it attaches raygun logging via: `RaygunWebApiClient.Attach(config)`

The problem is that the request body is not logged to raygun when an error is thrown.

In RaygunWebApiClient.cs, the attach method explicitly excludes RaygunWebApiDelegatingHandler when Owin assembly is detected. This handler is used to extract request body and to store it in request properties. The request body in the properties is then used in the message builder to generate the log.

```
var owinAssembly = AppDomain.CurrentDomain
    .GetAssemblies()
    .FirstOrDefault(a => a.FullName.StartsWith("Owin"));
if (owinAssembly == null)
{
    config.MessageHandlers.Add(new RaygunWebApiDelegatingHandler());
}
```

So, to simplify; it's not generating a log because the handler is been excluded. What's the reason for not adding the handler in owin applicationS? Is there a better way to log request body?

gregpakes:
I would like to know the answer to this too.  It looks like you won't get a request body if you have any assemblies with "owin" in their name.

Why is this?

QuantumNightmare:
Apologies for not replying to this. I remember the issue we ran into is if Owin is involved, the stream that provides the request body can only be read once. When Raygun was reading it to include the body in the report, the body could not be read again which caused problems in the web application itself. Checking for an Owin assembly was thought to be the easiest way to work around this.

I'd like us to revisit this but do not have a time frame of when we'd be able to. At the very least, I'd expect checking CanSeek on the Stream would work and be better than checking for an Owin assembly. In our AspNetCore provider, we still read unseekable streams and then copy the result to a new stream and put that back on the request so that Raygun can still get the body, and other things can still read it. Maybe we'd be able to do that here too.

gregpakes:
Thanks for the response.  Just a side note, I am not using AspNetCore (you mentioned the AspNetCore provider above)

gregpakes:
How quickly could you verify and accept a pull request matching the functionality to the AspNetCore provider?

I can potentially allocate some resources to submitting this.

xt0rted:
Here's my implementation for Raygun.Owin:

https://github.com/xt0rted/Raygun.Owin/blob/148835312e75c87da83490653f9d09e690ed2978/src/Raygun.Owin/Builders/RaygunRequestMessageBuilder.cs#L39-L44

[...]

PR #382, opened 2018-05-07
Added support for request body in projects with owin

gregpakes:
This is a PR for issue https://github.com/MindscapeHQ/raygun4net/issues/288.

This seems far too simple to be correct, but in all my testing it appears to work very well.  I have replicated the original issue with Owin and the issue no longer occurs with this new code.

The new code works because **ReadAsStringAsync** is always readable because it calls **LoadIntoBufferAsync** internally which then uses an internal buffer to allow re-reads.

This is the solution provided here: https://stackoverflow.com/a/13021877/584285


gregpakes:
I guess one issue here is the previous artificial limit of 4096 to the RawData that was imposed.  I'll see if I can replicate that now.

PR #391, opened 2018-08-31
Make RaygunClient implement IDisposable

ObeA:
RaygunClient initialises 3 instances of ThreadLocal<> which implements IDisposable. RaygunClient should dispose these when it runs out of scope. 

Also, I added a shell version of buildAspNetCore.ps1.


QuantumNightmare:
Hi ObeA, apologies for the delayed reply and thanks for the PR. I have added a task to our back log to review this. I do not have a time frame of when we'll be able to get on to this though.

PR #395, opened 2018-10-04
Fix StackOverflowException with .NET Core client

### LINKED ISSUE #392: StackOverflowException with .NET Core client
This took me a while to pin down, but the _SimpleJson.cs_ class used in Mindscape.Raygun4Net.NetCore.Common does not seem to be able to serialize `Type`s correctly - instead it throws a `StackOverflowException` and crashes the process.

Here's a simple repro:

```csharp
var data = new Dictionary<string, object> {{"Error", typeof(SimpleJson)}}; // userdata contains a Type

var message = RaygunMessageBuilder.New(new RaygunSettings { ApiKey = "" })
    .SetUserCustomData(data)
    .Build();

SimpleJson.SerializeObject(message); // throws StackOverflowExcepion and crashes the process
```

Maintaining your own JSON serializer seems a little pointless here, especially when you're including Newtonsoft.Json in Mindscape.Raygun4Net.NetCore anyway... It seems like using Newtonsoft.JSON would be the safest option. 

Alternatively, update SimpleJson to handle serializing `Type`s by calling `.AssemblyQualifiedName` on them, instead of trying to do a deep serialization (which I assume is what's calling the recurrsion)

j5alive:
It appears modifications had been made to the SimpleJson class to work with Net core, however these modifications weren't needed and broke the recursive serialization checks. The PR brings the implementation back in line with the version in the full framework project and resolves #392 

Other minor fixes
- Removed Newtonsoft.Json as a package reference.
- Fixed file references between projects to be project references and not nuget references.
- Fixed SimpleJson unit tests as these must have never been run.

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Common/SimpleJson.cs
+++ b/Mindscape.Raygun4Net.NetCore.Common/SimpleJson.cs
@@ -42,15 +42,15 @@
 // usually already defined in properties
 //#define NETFX_CORE;
 
-// If you are targetting WinStore, WP8, NET4.5+ PCL, or .NET Core make sure to #define SIMPLE_JSON_TYPEINFO;
+// If you are targetting WinStore, WP8 and NET4.5+ PCL make sure to #define SIMPLE_JSON_TYPEINFO;
 
-#define SIMPLE_JSON_TYPEINFO
+// #define SIMPLE_JSON_TYPEINFO
 
 // original json parsing code from http://techblog.procurios.nl/k/618/news/view/14605/14863/How-do-I-write-my-own-parser-for-JSON.html
 
-#if NETFX_CORE
+//#if NETSTANDARD1_6
 #define SIMPLE_JSON_TYPEINFO
-#endif
+//#endif
 
 using System;
 using System.CodeDom.Compiler;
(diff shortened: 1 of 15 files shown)


mduncan26:
I have a vague feeling I was the one who made the stupid changes to SimpleJson?

Mindscape.Raygun4Net.NetCore.Common/SimpleJson.cs
```diff
@@ -42,15 +42,15 @@
...
-// If you are targetting WinStore, WP8, NET4.5+ PCL, or .NET Core make sure to #define SIMPLE_JSON_TYPEINFO;
+// If you are targetting WinStore, WP8 and NET4.5+ PCL make sure to #define SIMPLE_JSON_TYPEINFO;
 
-#define SIMPLE_JSON_TYPEINFO
+// #define SIMPLE_JSON_TYPEINFO
 
 // original json parsing code from http://techblog.procurios.nl/k/618/news/view/14605/14863/How-do-I-write-my-own-parser-for-JSON.html
 
-#if NETFX_CORE
+//#if NETSTANDARD1_6
```
mduncan26: Are these guards needed if they are commented out?

j5alive: Oops, that was me testing, should not be commented out. netstandard20 includes the type support, so is really only needed for netstandard16 support

PR #397, opened 2018-11-19
[XRAY-2377] Release Xamarin Android with its own NuGet package.

mduncan26:
Building a separate NuGet package for this provider so it can be updated and used without referencing the main Raygun4Net package.

### DIFF
--- a/buildXamarinAndroid.sh
+++ b/buildXamarinAndroid.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+rm -rf Build/Xamarin/iOS
+# iOS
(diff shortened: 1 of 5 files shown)


buildXamarinAndroid.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+rm -rf Build/Xamarin/iOS
```
QuantumNightmare: throw new CopyAndPasteException("This should be Android, as well as the comment below.");

mduncan26: catch(CopyAndPasteException exception) {
  // Try again mitch.
  FindAndReplace("iOS", "Android");
}

PR #402, opened 2019-02-12
[XRAY-2615] Xamarin iOS Better Offline Storage Support

mduncan26:
This PR looks to address task exception that occur when sending stored reports.

- Port offline support logic from Xamarin Android
- Include unobserved task exception handling logic

### DIFF
--- a/Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunFileManager.cs
+++ b/Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunFileManager.cs
@@ -66,7 +66,7 @@ namespace Mindscape.Raygun4Net
       // Write data to file path.
       using (var writer = File.Create(filePath))
       {
-        var bytes = Encoding.ASCII.GetBytes(data);
+        var bytes = Encoding.UTF8.GetBytes(data);
         writer.Write(bytes, 0, bytes.Length);
       }
 
--- a/Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunFileManager.cs
+++ b/Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunFileManager.cs
@@ -1,0 +1,185 @@
+﻿using System.Text;
+using System.Collections.Generic;
+using Mindscape.Raygun4Net.Messages;
+using System;
+using System.IO;
+
+namespace Mindscape.Raygun4Net
+{
+  public class RaygunFileManager
+  {
+    public const int MAX_STORED_REPORTS_UPPER_LIMIT = 64;
+    private const string RAYGUN_DIRECTORY = "RaygunIO";
+
+    private int currentFileCounter = 0;
+
+    public void Intialise()
+    {
+      try
+      {
+        var crashReportDirectory = GetCrashReportDirectory();
+
+        if (!Directory.Exists(crashReportDirectory))
+        {
+          Directory.CreateDirectory(crashReportDirectory);
+        }
+      }
+      catch (Exception e)
+      {
+        RaygunLogger.Error("Failed to initialise file manager due to error: " + e.Message);
+      }
+    }
+
+    public string SaveCrashReport(RaygunMessage crashReport, int maxReportsStored)
+    {
+      try
+      {
+        // Can only save the report if we havnt reached the report count limit.
+        if (IsFileLimitReached(Math.Min(maxReportsStored, MAX_STORED_REPORTS_UPPER_LIMIT)))
+        {
+          RaygunLogger.Warning("Failed to store crash report - Reached max crash reports stored on device");
+          return null;
+        }
+
+        // Convert to JSON text.
+        var reportJson = SimpleJson.SerializeObject(crashReport);
+
+        // Generate a unique name for our report.
+        var fileName = GetUniqueAcendingJsonName();
+
+        // Save the report to disk.
+        return StoreCrashReport(fileName, reportJson);
+      }
+      catch (Exception ex)
+      {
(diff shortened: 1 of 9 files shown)


Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunResponseSatusCode.cs
```diff
@@ -0,0 +1,43 @@
...
+        case (int)RaygunResponseStatusCode.Accepted:
+          return "(202) Request succeeded";
+
+        case (int)RaygunResponseStatusCode.BadMessage:
+          return "(400) Bad message - Could not parse the provided JSON.";
+
+        case (int)RaygunResponseStatusCode.InvalidApiKey:
+          return "(403) Invalid API Key - The value specified in the header X-ApiKey did not match with an application in Raygun";
+
+        case (int)RaygunResponseStatusCode.RequestTimeout:
```
j5alive: Probably more likely to get a 504 timeout (Gateway timeout) so could be good to account for that too.

Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunFileManager.cs
```diff
@@ -0,0 +1,185 @@
...
+
+    private string StoreCrashReport(string fileName, string data)
+    {
+      // Generate the full file path.
+      var filePath = Path.Combine(GetCrashReportDirectory(), fileName);
+
+      // Write data to file path.
+      using (var writer = File.Create(filePath))
+      {
+        var bytes = Encoding.ASCII.GetBytes(data);
```
j5alive: You may want to use UTF8 encoding here to match the sending of this Data in the client.

Mindscape.Raygun4Net.Xamarin.iOS/RaygunClient.cs
```diff
@@ -538,6 +471,240 @@ private static void CurrentDomain_UnhandledException(object sender, UnhandledExc
...
+    private void SendAllStoredCrashReports()
+    {
+      if (!HasInternetConnection)
+      {
+        RaygunLogger.Debug("Not sending stored crash reports due to no internet connection");
+        return;
+      }
+
+      // Get all stored crash reports.
+      var reports = _fileManager.GetAllStoredCrashReports();
```
QuantumNightmare: Could bringing all saved reports into memory all at once be an issue? For a max of 64 files, _maybe_ that's fine, but if we were to remove/bump the max file limit in future and not revise this logic, I feel that it could become an issue.

Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunFileManager.cs
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    public string SaveCrashReport(RaygunMessage crashReport, int maxReportsStored)
+    {
+      try
+      {
+        // Can only save the report if we havnt reached the report count limit.
+        if (IsFileLimitReached(Math.Min(maxReportsStored, MAX_STORED_REPORTS_UPPER_LIMIT)))
+        {
+          RaygunLogger.Warning("Failed to store crash report - Reached max crash reports stored on device");
```
QuantumNightmare: We used to start overriding the oldest saved report when reaching the maximum file limit. Looks like we've lost that here, so questioning if this is really what we want. The most recent reports are likely more useful than stale ones.

Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunFileManager.cs
```diff
@@ -0,0 +1,185 @@
...
+        var crashReportDirectory = GetCrashReportDirectory();
+
+        if (!Directory.Exists(crashReportDirectory))
+        {
+          Directory.CreateDirectory(crashReportDirectory);
+        }
+      }
+      catch (Exception e)
+      {
+        RaygunLogger.Error("Failed to initialise file manager due to error: " + e.Message);
```
QuantumNightmare: Minor: I notice we use string.Format and concatenation over string interpolation. Just wondering any reason behind this.

mduncan26: Got lazy writing out string.format. Ive changed this in the new xamarin forms provider to use a params list and format them just before printing them.

Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunFileManager.cs
mduncan26: Yeah, this is a point I have discussed with Jerms on a previous PR. We found this to be a better solution to avoid losing out on error reports that are overwritten by a flood of one noisy error exception. It also means less file management and deleting of files, reducing the chance of causing errors ourself.

Mindscape.Raygun4Net.Xamarin.iOS/RaygunClient.cs
mduncan26: Thats a fair point. Ill look to address this in the new Xamarin Forms provider.

Mindscape.Raygun4Net.Xamarin.iOS.Unified/RaygunFileManager.cs
QuantumNightmare: :+1:

Mindscape.Raygun4Net.Xamarin.iOS/RaygunClient.cs
QuantumNightmare: :+1:

PR #426, opened 2020-02-13
Use Static HttpClient in .NetCore Projects

MattByers:
# Overview 🚉 
Previously the .Net Core projects were new'ing up an HttpClient on every Send call. This was potentially causing connection pool exhaustion. The HttpClient class is designed to be a single instance that will manage it's connections internally, therefore a static instance makes sense. There could be issues around out of date DNS changes, so long term we will look to use an HttpClientFactory or something similar.

# Changes 🔧 
- Create static HttpClient in RaygunClientBase
- Use static client instead of new'd up client in using statement


phillip-haydon:
This doesn't appear to be needed anymore, there is now a DefaultClient and the ability to supply a HttpClient via ctor.

PR #429, opened 2020-06-12
Add SendInBackground overload to NetCore projects to support sending an Exception with a RaygunIdentifierMessage

### LINKED ISSUE #428: Consider adding Raygun4Net's exception + userInfo SendInBackground overload to .NetCore package
Hey there!

We're migrating from `Mindscape.Raygun4Net` to `Mindscape.Raygun4Net.NetCore`, however we've discovered that the overload to send an `exception` with `userInfo` is not available in the .NetCore package, as it is in the Raygun4Net package.

The source in the original package is here:

https://github.com/MindscapeHQ/raygun4net/blob/637422bc0c28a06f3dda26737f118a0c632a019a/Mindscape.Raygun4Net4.ClientProfile/RaygunClient.cs#L185

We use the same `RaygunClient` concurrently across multiple error reporting requests, and as such would prefer to use this overload instead of setting the `UserInfo` on `RaygunClient` before sending the request.

I'd be happy to open a PR with this functionality if it's something that you're open to considering. Cheers!

josh-:
Resolves #428.

Adds a `SendInBackground` overload to the NetCore packages which supports sending an `Exception` with a `RaygunIdentifierMessage`.

This mirrors what is currently available in the `Mindscape.Raygun4Net` package:

https://github.com/MindscapeHQ/raygun4net/blob/d59749c03519a5210409408a7aa2bb1a122c5e25/Mindscape.Raygun4Net/RaygunClient.cs#L289-L296

### DIFF
--- a/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
+++ b/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
@@ -241,7 +241,7 @@ namespace Mindscape.Raygun4Net.AspNetCore
             {
                 _currentRequestMessage.Value = currentRequestMessage;
                 _currentResponseMessage.Value = currentResponseMessage;
-                await StripAndSend(exception, tags, userCustomData, null);
+                await StripAndSend(exception, tags, userCustomData, userInfo);
             });
             FlagAsSent(exception);
             await task;
--- a/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
+++ b/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
@@ -227,17 +227,6 @@ namespace Mindscape.Raygun4Net.AspNetCore
     /// <param name="exception">The exception to deliver.</param>
     /// <param name="tags">A list of strings associated with the message.</param>
     /// <param name="userCustomData">A key-value collection of custom data that will be added to the payload.</param>
-    public Task SendInBackground(Exception exception, IList<string> tags, IDictionary userCustomData)
-    {
-        return SendInBackground(exception, tags, userCustomData, null);
-    }
-
-    /// <summary>
-    /// Asynchronously transmits an exception to Raygun.
-    /// </summary>
-    /// <param name="exception">The exception to deliver.</param>
-    /// <param name="tags">A list of strings associated with the message.</param>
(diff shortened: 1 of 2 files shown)


Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
```diff
@@ -227,24 +227,36 @@ protected override bool CanSend(RaygunMessage message)
     /// <param name="exception">The exception to deliver.</param>
     /// <param name="tags">A list of strings associated with the message.</param>
     /// <param name="userCustomData">A key-value collection of custom data that will be added to the payload.</param>
-    public override async Task SendInBackground(Exception exception, IList<string> tags, IDictionary userCustomData)
+    public Task SendInBackground(Exception exception, IList<string> tags, IDictionary userCustomData)
```
mduncan26, line 230: This method can be removed since it is declared in the base class.

Compiler: RaygunClient.cs(230, 17): [CS0108] 'RaygunClient.SendInBackground(Exception, IList<string>, IDictionary)' hides inherited member 'RaygunClientBase.SendInBackground(Exception, IList<string>, IDictionary)'. Use the new keyword if hiding was intended.

Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
```diff
@@ -227,24 +227,36 @@ protected override bool CanSend(RaygunMessage message)
...
+            // We need to process the Request on the current thread,
+            // otherwise it will be disposed while we are using it on the other thread.
+            RaygunRequestMessage currentRequestMessage = await BuildRequestMessage();
+            RaygunResponseMessage currentResponseMessage = BuildResponseMessage();
+
+            var task = Task.Run(async () =>
+            {
+                _currentRequestMessage.Value = currentRequestMessage;
+                _currentResponseMessage.Value = currentResponseMessage;
+                await StripAndSend(exception, tags, userCustomData, null);
```
mduncan26, line 255: It appears you missed passing through the userInfo parameter to the `StripAndSend` method?

josh-:
Good pickups @mduncan26! Thanks, those should both be resolved now 😊

josh-:
Thanks so much @mduncan26 – really appreciate it! 🎉

mduncan26:
Hi @josh- these changes are now available through NuGet in version 6.3.0 of the package.

PR #434, opened 2020-07-17
[CR-204] Have the SendingMessage event include the original exception 

mduncan26:
Update the `SendingMessage` event to include the original exception alongside the `RaygunMessage` object. This allows customers to perform further operations on the exception to extract contextual information only available on the exception.

Ticket: https://raygun.atlassian.net/browse/CR-204

### DIFF
--- a/AssemblyVersionInfo.cs
+++ b/AssemblyVersionInfo.cs
@@ -10,5 +10,5 @@ using System.Reflection;
 // You can specify all the values or you can default the Build and Revision Numbers
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("5.10.2.0")]
-[assembly: AssemblyFileVersion("5.10.2.0")]
+[assembly: AssemblyVersion("5.11.0.0")]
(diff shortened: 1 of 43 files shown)


samuel-holt (dismissed):
:+1: This change seems straightforward, nice work

MattByers (dismissed):

QuantumNightmare (changes requested):
I have pointed out some methods that now require the new Exception parameter as they don't specify a default value, and I believe this would cause a breaking change.

I'm also going to point out this [somewhat related blog post](https://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/) around being careful when using optional parameters. It suggests that adding optional parameters to an existing method can cause runtime exceptions if an application has not been compiled, but has a new version of a client library that has added optional parameters to existing methods. I would expect applications to typically be compiled when upgrading to newer versions of client libraries though, so maybe it's not a concern for us.

phillip-haydon:
Superseded by `IMessageBuilder`, which provides access to the exception and can flag the resulting message for cancellation before it is sent.

PR #439, opened 2020-09-30
Expose BuildStackTrace function

QuantumNightmare:
This PR exposes a function on the ErrorMessageBuilder to build an array of stacktrace line models from a given .NET StackTrace instance. This is to share the code with the Serilog Raygun Sink in this PR: https://github.com/serilog/serilog-sinks-raygun/pull/30

Additionally I've removed the Xamarin Component files (except the samples as they may still be useful).


mduncan26 (approved):
LGTM

Issue #421, opened 2019-12-10
HTTP request context is lost after building first RaygunRequestMessage


Zero3:
From the source:

````
    private RaygunRequestMessage BuildRequestMessage()
    {
      var message = _currentWebRequest.Value != null ? RaygunWebApiRequestMessageBuilder.Build(_currentWebRequest.Value, _requestMessageOptions) : null;
      _currentWebRequest.Value = null;
      return message;
    }
````

This means that every time a `RaygunRequestMessage` is built by `Send()` or `SendInBackground()`, the current HTTP request context is lost when this code sets `_currentWebRequest` to `null` for the current thread.

The implication is that each subsequent error sent by a `RaygunWebApiClient` instance won't contain all the information that normally is extracted from the HTTP request context.

Perhaps a `RaygunWebApiClient` is only intended to be used to send one error? If so, that is not very clear from the code or documentation, and also kind of contradicts `README.md` (highlighting by me):

> Sometimes when setting up Raygun to send exceptions automatically, you may need to provide a custom RaygunWebApiClient instance in order to use some of the optional feature described below.
To do this, use the static RaygunWebApiClient.Attach method overload that takes a function. Within this function, return a new (**or previously created**) RaygunWebApiClient instance.

knumat:
This bug still exists.
https://github.com/MindscapeHQ/raygun4net/blob/fefc40dc0e5d40691b968c6fa69ab60184457293/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs#L254

Additionally, this bug also means that Message.Details.Response is never populated, since BuildResponseMessage() is called after BuildRequestMessage() sets  _currentHttpContext.Value to null.

What do we need to get this fixed?  Should I submit a pull request?

mduncan26:
Hi @knumat 

Yes, we are open to pull requests. Otherwise I will raise this issue with the team and we will look to create an internal ticket to address this issue. I am unable to give an accurate time in which this issue will be addressed. However as this is a bug within the provider we will look to prioritise it higher. If you wish to submit a PR for this issue that will help us reduce the time it takes to get an official release out. We will update this thread if there is any change to the progress of addressing this issue.

Thank you,
Mitchell.

knumat:
As best I can tell, this bug appears to have been added by the following commit.
[...]

PR #440, opened 2020-10-28
#421 No need to clear the current HttpContext when populating

knumat:
This caused a subtle bug where the first exception reported would include the request context, but subsequent exceptions would not.

In addition, the AspNetCore version would never set the StatusCode or StatusDescription in the Response
since _currentHttpContext was set to null by BuildRequestMessage() just before BuildResponseMessage() was called.
Note that StatusCode would not typically be set if an exception occurred on ASP.NET Core since the Raygun HTTP
module does not exist on this platform, but custom middleware might want to do this.
For example, custom middleware might watch for certain HTTP status codes and report them.

As long as RaygunClient objects are short-lived, this should not cause any memory leaks.

### DIFF
--- a/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
+++ b/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
@@ -250,16 +250,12 @@ namespace Mindscape.Raygun4Net.AspNetCore
 
     private async Task<RaygunRequestMessage> BuildRequestMessage()
     {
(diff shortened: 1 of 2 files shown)


mduncan26:
Hi @knumat 

Thank you for the PR we will updater this thread once we have had a chance to review the changes.

Regards,
Mitchell.

phillip-haydon:
This will be resolved in #518 by using IHttpContextAccessor and passing the info in rather than attempting to store the data locally in the RaygunClient (RaygunClient should be treated as a singleton)

Issue #371, opened 2018-01-01
SendInBackground should not use ThreadStatic


snakefoot:
https://github.com/MindscapeHQ/raygun4net/blob/0f232cd9829bcda2d769ae7a26249a149a041d12/Mindscape.Raygun4Net4/RaygunClient.cs#L335

Should avoid populating ThreadStatic variables for ThreadPool-Threads.

Passing parameters using global variables is generally a bad idea.

knumat:
This is a problem, since the thread that performs the async operation may not be the same thread that queued the operation.  If not, the variables will be null, and the error message will be sent without the expected context.

Note that some versions of the library use ThreadLocal instead of ThreadStatic, but that doesn't fix this potential issue.  For example:
https://github.com/MindscapeHQ/raygun4net/blob/fefc40dc0e5d40691b968c6fa69ab60184457293/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs#L17

For async-based operations, you should use AsyncLocal instead of ThreadLocal, as shown in the following link.  I don't know what the equivalent is in the older style of async programming that uses ThreadPool.QueueUserWorkItem().
https://docs.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=netcore-3.1

Let me know if a pull request would be helpful.

mduncan26:
Hi @knumat 

Thank you for the additional information. We definitely are open to pull requests. Please feel free to open pull request for any additional you feel you can provide solutions to. We really do appreciate the community helping resolve these outstanding problems. I will also raise this issue with the team and we will look to create an internal ticket to address this issue. We will update this thread if there is any change to the progress of addressing this issue.

Thank you,
Mitchell.

knumat:
I created a PR with a couple items, but I think there is still an over-arching issue to resolve.  While the ASP.NET Core version might be fixed by the PR, the remaining versions still have a potential issue.  If one thread starts an operation, and a different thread continues the operation, the ThreadStatic/ThreadLocal variables will be null, leading to possible data loss.  I'm not familiar enough with each platform's scheduler to know how much of an issue this might be.

Obi-Dann:
Hello, is there any update on it? There's still an open PR awaiting for a review. And it's still an issue that ThreadLocal is used instead of AsyncLocal

phillip-haydon:
[...]

PR #441, opened 2020-10-28
Use async local

knumat:
Here are a few cleanups related to #371.

The ASP.NET Core version was updated from ThreadLocal to AsyncLocal, since it uses async-based code:
https://docs.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=netcore-3.1#examples

Additionally, I removed an initializer for a ThreadStatic variable, since initializers are only applied to the first thread that accesses the ThreadStatic.  The variable for each subsequent thread uses the default initializer.  Another option would be to replace ThreadStatic with ThreadLocal, which allows you to provide an initializer, but that wasn't necessary in this case.
https://docs.microsoft.com/en-us/dotnet/api/system.threadstaticattribute?view=netcore-3.1#remarks

### DIFF
--- a/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
+++ b/Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
@@ -14,9 +14,9 @@ namespace Mindscape.Raygun4Net.AspNetCore
   {
     protected readonly RaygunRequestMessageOptions _requestMessageOptions = new RaygunRequestMessageOptions();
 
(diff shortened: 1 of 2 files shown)


mduncan26:
Hi @knumat 

Thank you for the PR we will updater this thread once we have had a chance to review the changes.

Regards,
Mitchell.

phillip-haydon:
In #518 we have removed the usage of ThreadLocal entirely.

PR #442, opened 2020-11-12
Add an override for the SendAsync method in the ASP.NET CORE RaygunClient

samuel-holt:
## Overview

When using the `RaygunClient.Send()` overload in an ASP.NET Core MVC application, the HTTP request details are not being included in the message being sent to the API. This change adds an override for the `SendAsync` method which awaits the retrieval of the request details and sets them on the ThreadLocal value which is used in the `BuildMessage` method.

### Updates

- Add override for the SendAsync method in the ASP.NET Core RaygunClient 
- The HTTP context is not set to null until after the request and response messages have been built
- Add docs to the readme for the case where a singleton RaygunClient object is used

### DIFF
--- a/Mindscape.Raygun4Net.AspNetCore/README.md
+++ b/Mindscape.Raygun4Net.AspNetCore/README.md
@@ -127,35 +127,19 @@ public class RaygunController : Controller
 
 ### Using a singleton RaygunClient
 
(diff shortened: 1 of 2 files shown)


QuantumNightmare (approved):
Looks good :+1:

QuantumNightmare (approved):
That's better, thanks :+1:

PR #444, opened 2020-12-07
Update psake

QuantumNightmare:
Updated the build scripts to use the latest version of psake. This fixes a path issue with finding msbuild in VS 2019.
Also updated the version number of the Raygun4Net.Mvc project which I'd missed in a previous piece of work.


mduncan26 (approved):
LGTM

PR #447, opened 2021-02-04
Fix to the key value pair data filter

mduncan26:
Changes:
- fix an exception when accessing string array out of bounds


MattByers (approved):
🥇

PR #449, opened 2021-03-30
[APM-475] Remove APM profiler support from Raygun4Net.AspNetCore

Deacon-McIntyre:
Ticket: [https://raygun.atlassian.net/browse/APM-475](https://raygun.atlassian.net/browse/APM-475)

Raygun4Net will take over the APM sampling from the agent if Raygun4Net is installed into an app that has the APM profiler attached. This works for .NET Framework but will not work for .NET Core due to .NET Core’s asynchronous nature. In fact it will result in many traces being lost.

We are removing Raygun4Net’s sampling integration so that traces will behave as normal even when the application has Raygun4Net installed, because customers should be able to use APM and CR in the same application.

### Testing
I added the Raygun4Net.ASpNetcore project to into the solution of a .NET Core test project, and manually sent an exception.

I also checked the sampling overrides are applied as expected while Raygun4Net was installed.


mduncan26 (approved):
Looks good, don't forget you will need to update the package version.

mduncan26 (approved):
LGTM

PR #450, opened 2021-04-19
[APM-488] Disable profiling support

Deacon-McIntyre:
Raygun4Net will take over the APM sampling from the agent if Raygun4Net is installed into an app that has the APM profiler attached. We have removed this for .NET Core - see https://github.com/MindscapeHQ/raygun4net/pull/449

Since it has been removed for the .NET Core part of Raygun4Net, we are opting to also remove it from the rest of Raygun4Net. This is for both consistency and maintainability.

**Testing**
* [x] Test that profiler sampling has affect before these changes
* [x] Build project(s) and check version has been updated
* [x] Reference locally built packages
* [x] Check that profiler sampling has no effect
* [x] Check that errors still come through


AssemblyVersionInfo.cs
```diff
@@ -10,5 +10,5 @@
 // You can specify all the values or you can default the Build and Revision Numbers
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("5.12.1.0")]
```
mduncan26, line 13: Lets revert these changes since we split out the libraries to have their own copy of this file.

PR #455, opened 2021-08-23
Add ConfigureAwait(false)

ceilingfish:
Add ConfigureAwait(false) to library calls to allow correct operation in certain .Net Framework scenarios. (e.g. WinForms & WebForms).


JasenPalmer (approved):
Thank you for making these changes. We will look to get this update published shortly.

PR #458, opened 2021-12-22
Fix missing cookie collection in ASP.NET Core provider

j5alive:
The check for `request.HasFormContentType` when trying to get cookies from the request should not be there as the Form content has no bearing on a request containing cookies or not. It looks like this was copied over from the original implementation.

I've also added a message to the cookies list if this method fails in line with the other methods in this class.


j5alive:
@QuantumNightmare any chance you could review this please?

QuantumNightmare (approved):
Looks good, apologies for the wait.

QuantumNightmare:
This fix has been released in [version 6.6.5 of the NuGet package](https://www.nuget.org/packages/Mindscape.Raygun4Net.AspNetCore/6.6.5) now.

PR #461, opened 2022-02-15
Darcy/apl 188

darcythomas:


JasenPalmer (approved):
Looks good to me!

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -48,6 +48,44 @@ public abstract class RaygunClientBase
...
+          System.AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
+          _settings.CatchUnhandledExceptions = true;
+          return;
+        }
+        else
+        {
+          System.AppDomain.CurrentDomain.UnhandledException -= CurrentDomain_UnhandledException;
+          _settings.CatchUnhandledExceptions = false;
+          return;
+        }
```
JasenPalmer, line 78: Not sure if this is your style but you can remove the `return` statements and also could set `_settings.CatchUnhandledExceptions = value;` after the if/else block

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -59,8 +97,16 @@ public RaygunClientBase(RaygunSettingsBase settings)
       {
         ApplicationVersion = settings.ApplicationVersion;
       }
+
+#if NETSTANDARD2_0_OR_GREATER
+            if (_settings.CatchUnhandledExceptions)
+            {
+               CatchUnhandledExceptions = _settings.CatchUnhandledExceptions;
+            }
```
JasenPalmer, line 105: Indentation is a bit off here

JasenPalmer (approved):
Looks good to me. Just a couple of minor comments to checkout

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
darcythomas: The `_settings.CatchUnhandledExceptions = value`  needs to come first or otherwise you could add more than one event handler, which would be bad!

JasenPalmer (approved):
Thanks for the changes. Looks good

PR #467, opened 2022-04-07
Initialize Tags as List<string> instead of string array in ASP.NET Core middleware

### LINKED ISSUE #466: NotSupportedException when adding a tag from ASP.NET middleware
When my application throws an exception that is logged by Raygun, I get a second exception report from the Raygun client itself that occurs when my Raygun Client Provider tries to add a tag:

```csharp
public class RaygunClientProvider : DefaultRaygunAspNetCoreClientProvider
{
    public override RaygunClient GetClient(RaygunSettings settings, HttpContext context)
    {
        var client = base.GetClient(settings, context);

        client.SendingMessage += (_, args) =>
        {
            args.Message.Details.Tags ??= new List<string>();
            args.Message.Details.Tags.Add(Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT"));
        };

        return client;
    }
}
```

The call to `Tags.Add()` throws `System.NotSupportedException`: "Collection was of a fixed size." The stack trace from the Raygun report:

```json
{
  "Error": {
    "Data": {},
    "ClassName": "System.NotSupportedException",
    "Message": "Collection was of a fixed size.",
    "StackTrace": [
      {
        "LineNumber": 11,
        "ClassName": "System.ThrowHelper",
        "MethodName": "ThrowNotSupportedException(ExceptionResource resource)"
      },
      {
        "LineNumber": 0,
        "ClassName": "System.SZArrayHelper",
        "MethodName": "Add[T](T value)"
      },
      {
        "LineNumber": 17,
        "ClassName": "WebApp.Platform.Raygun.RaygunClientProvider+<>c",
        "FileName": "D:\\projects\\reports\\src\\WebApp\\Platform\\Raygun\\RaygunClientProvider.cs",
        "MethodName": "<GetClient>b__0_0(Object _, RaygunSendingMessageEventArgs args)"
      },
      {
        "LineNumber": 27,
        "ClassName": "Mindscape.Raygun4Net.RaygunClientBase",
        "MethodName": "OnSendingMessage(RaygunMessage raygunMessage)"
      }
    ]
  }
}
```

I have another, very similar application with the same Client Provider where this issue does not occur. The only relevant difference between them that I can think of is that this app has `<Nullable>enable</Nullable>` in the "csproj" file. Could it be that with nullable reference types enabled, the `Tags` property is getting initialized with a fixed-size data structure, like an Array?

dougwaldron:
When the UnhandledException tag is added by the ASP.NET Core middleware, tags are initialized as a string array, which
is of fixed size, preventing additional tags from being added (for example when overriding the Raygun client provider). Initializing as a `List<string>` instead will fix this.

fixes #466

### DIFF
--- a/Mindscape.Raygun4Net.AspNetCore/RaygunAspNetMiddleware.cs
+++ b/Mindscape.Raygun4Net.AspNetCore/RaygunAspNetMiddleware.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.IO;
@@ -81,7 +82,7 @@ namespace Mindscape.Raygun4Net.AspNetCore
         }
 
         var client = _middlewareSettings.ClientProvider.GetClient(_settings, httpContext);
-        await client.SendInBackground(e, new []{UnhandledExceptionTag});
+        await client.SendInBackground(e, new List<string>{UnhandledExceptionTag});
         throw;
       }
       finally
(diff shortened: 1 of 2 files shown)


darcythomas (approved):
Looks good thanks

PR #476, opened 2023-05-16
[MC-88] README changes for ASP.NET

crystalqdm:
This PR is to update the README to distinguish installation instructions for ASP.NET 5+ (Also known and ASP.NET Core) and ASP.NET Framework


Hamish-taylor (approved):
Just one suggestion, feel free to ignore it if you don't think it makes sense. Otherwise it looks good to me!

PR #477, opened 2023-05-17
Updating WebClient usage for better TLS handling  

darcythomas:
Background: 

TLS 1.1 (+) does not work out of the box on .net framework 4.0 or 4.6.

This fix sets the tls version to use, when sending a event to RG.

After sending it resets the config back to the previous settings so we dont let out library (Raygun4net) force a change in a app wide global variable.

I have tested this on my machine (win 10) against [net4 net45 net462 net47 net471 net472 net48]

I also tried to get it to work with .net 6 but this proved too difficult, for now, and the code was removed from this PR

I have added a number of unit tests, so of which are temporary (`TestCurrentTls.cs`).
 I wanted to include them in the PR so the team can see what we started with.
Before the final merg I will remove them.

I have added a new project `Mindscape.Raygun4Net.Common\Mindscape.Raygun4Net.Common.csproj` because I needed some to have some common files in a new sdk compatible project.
 I am hoping that as time allows we will be able to move items from `Mindscape.Raygun4Net.Core\Mindscape.Raygun4Net.Core` (which does not have anything to do with .net core) into this new project.

I will create a ticket and another PR for those changes.

I have noticed some **sub-optimal** use of async. Ref:
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md
(Causing extra thread usage and possibly deadlocks)

I will create a ticket and another PR for those changes.


Mindscape.Raygun4Net.Common/DataAccess/WebClientFacade.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    public WebClientFacade(WebClient client)
+    {
+      _client = client;
+    }
+
+    public string UploadString(Uri address, string data)
+    {
+
+#if NET45
```
Olwiba, line 20: nit: should these be in order?

Mindscape.Raygun4Net.Common/DataAccess/WebClientFacade.cs
```diff
@@ -0,0 +1,42 @@
...
+      var holdingPen = ServicePointManager.SecurityProtocol;
```
Olwiba, line 21: nit: var name, would something like `currentProtocol` better represent the previous TLS version

Olwiba (commented):
No major issues with the code changes here. Loved the tests!
As there are others on this PR for review, I've leave approving this for them

Mindscape.Raygun4Net.Common/DataAccess/WebClientFacade.cs
darcythomas: Yes.
But have changed things up a bit.

Mindscape.Raygun4Net.Common/DataAccess/WebClientFacade.cs
darcythomas: Yup, but have changed this any way.

Olwiba (commented):
Thanks for the changes :D

xenolightning (dismissed):
Looking really good, and nice work on the test coverage 😎

Mindscape.Raygun4Net4.Intergration.Tests/TestIfWeCanSendToRaygun.cs
```diff
@@ -0,0 +1,79 @@
...
+    [TearDown]
+    public void EndTest()
+    {
+      Trace.Flush();
+    }
+
+    /// <summary>
+    /// This checks that a local user secret has been set for this project
+    ///
+    /// Make your you set the api ke o a real key, but don't commit it!
```
QuantumNightmare, line 35: bit of a typo here

Mindscape.Raygun4Net4.Intergration.Tests/TestIfWeCanSendToRaygun.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// </summary>
+    [Test]
+    public void CheckApiKeyHasBeenSet()
+    {
+      var sut = new RaygunClient(ApiKey);
+      sut.Send(new NotImplementedException(nameof(CanSendWithValidKey)));
+
+      Assert.IsNotEmpty(ApiKey);
+      Assert.IsNull(TraceChecker.Traces.FirstOrDefault(f=> f.Equals("Raygun: Failed to send offline reports due to invalid API key.")));
```
QuantumNightmare, line 44: Might be a bit fragile if the wording of that message changes, then this will always pass.

Mindscape.Raygun4Net4.Intergration.Tests/TestIfWeCanSendToRaygun.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    [TestCase(ApiKey)]
+    [TestCase(InvalidApiKey)]
+    public void CanTrySendAndHasNoSecureChannelErrors(string apiKey)
+    {
+      var sut = new RaygunClient(apiKey);
+
+      sut.Send(new NotImplementedException(nameof(CanSendWithInvalidKeyButGetsA403)));
+
+      Assert.IsFalse(TraceChecker.Traces.Contains("Raygun: Failed to send report to Raygun due to: The request was aborted: Could not create SSL/TLS secure channel."));
```
QuantumNightmare, line 76: Again, this could always pass if the wording of the message changes. Could optionally strengthen this by making a constant.

Mindscape.Raygun4Net4/Mindscape.Raygun4Net4.csproj
```diff
@@ -42,14 +42,14 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Microsoft.VisualBasic" />
+    <!-- <Reference Include="Microsoft.VisualBasic" /> -->
```
QuantumNightmare, line 45: Could delete these commented lines if we don't need them

Mindscape.Raygun4Net4.Intergration.Tests/TestIfWeCanSendToRaygun.cs
darcythomas: ooh I like consts

Mindscape.Raygun4Net.Common/DataAccess/RaygunWebClient.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    private static readonly PropertyInfo SslProtocolsPropertyInfo =
+      typeof(HttpWebRequest).GetProperty("SslProtocols", BindingFlags.Instance | BindingFlags.NonPublic);
+
+    public RaygunWebClient(SecurityProtocolType? sslProtocol = null)
+    {
+      SecurityProtocolType defaults = ServicePointManager.SecurityProtocol;
+#if NET40
+   // defaults |= (SecurityProtocolType)768;//TLS 1.1
+    defaults |= (SecurityProtocolType)3072; //TLS 1.2
```
QuantumNightmare, line 23: nit: indentation a little off here

Mindscape.Raygun4Net4/WebClientHelper.cs
```diff
@@ -1,14 +1,25 @@
-using System;
+﻿using System;
 using System.Net;
+using System.Reflection;
+using Mindscape.Raygun4Net.Common.DataAccess;
 using Mindscape.Raygun4Net.Logging;
 
 namespace Mindscape.Raygun4Net
 {
+
```
QuantumNightmare, line 9: nit: lot's of extra new lines throughout this file

Mindscape.Raygun4Net4.Intergration.Tests/TestIfWeCanSendToRaygun.cs
darcythomas: So I would like to change the above! Gowver I dont think I can fic right noew because: 

1. But the message test comes from an exception.Message field. Which in these cases is reasonable stable.  
2. To do a proper fix I would need to change the code to be catching specific exceptions types. Which feels a bit out of scope for this PR. 
3. These tests are kinda tailored for dev time testing (needing to add an API key).


So do you mind if I leave them for now and look at them again as part of my 2-3.5 code removal PR?

buildSigned.ps1
```diff
@@ -40,7 +40,7 @@ task Compile -depends Init {
     exec { msbuild "$solution_file_net2" /m /p:OutDir=$build_dir_net2 /p:Configuration=$configuration }
     exec { msbuild "$solution_file_net35" /m /p:OutDir=$build_dir_net35 /p:Configuration=$configuration }
     exec { msbuild "$solution_file_net35_client" /m /p:OutDir=$build_dir_net35_client /p:Configuration=$configuration }
-    exec { msbuild "$solution_file_net4" /m /p:OutDir=$build_dir_net4 /p:Configuration=$configuration }
+    exec { "msbuild.exe $solution_file_net4 /m /p:OutDir=$build_dir_net4 /p:Configuration=$configuration" }
```
darcythomas, line 43: It would not build unless I did this. Appears to be an issue with psake https://github.com/psake/psake/issues/312

buildSigned.ps1
```diff
@@ -25,7 +25,7 @@ properties {
     $env:Path +=                  ";$nuget_dir"
 }
 
-task default -depends Compile, CompileWinRT
+task default -depends Compile #, CompileWinRT
```
darcythomas, line 28: Win RT fails to even open in VS on my machine. But I don't need it for this PR

Mindscape.Raygun4Net.Common/DataAccess/RaygunWebClient.cs
```diff
@@ -0,0 +1,46 @@
...
+    private static readonly PropertyInfo SslProtocolsPropertyInfo =
+      typeof(HttpWebRequest).GetProperty("SslProtocols", BindingFlags.Instance | BindingFlags.NonPublic);
+#endif
+
+    public RaygunWebClient(SecurityProtocolType? sslProtocol = null)
+    {
+      SecurityProtocolType defaults = ServicePointManager.SecurityProtocol;
+#if NET40
+      defaults |= (SecurityProtocolType)3072; //TLS 1.2
+      defaults |= (SecurityProtocolType) 12288; //TLS 1.3
```
darcythomas, line 24: Added this bonus future proofing.

README.md
```diff
@@ -8,7 +8,7 @@ Supported platforms/frameworks
 
 Projects built with the following frameworks are supported:
 
-* .NET 2.0, 3.5, 4.0+, up to .NET 7.0
+* .NET 4.0+, up to .NET 7.0
 * .NET 3.5 and 4.0 Client Profile
```
xenolightning, line 12: Are we still supporting .NET35 Client Profile? Is it a different target to the others?

buildSigned.ps1
xenolightning: Oh interesting! Does this pick up the msbuild in the path? And therefore use 2022 in this case?

README.md
darcythomas: oops that change should be in my other PR

darcythomas: So with this PR .net 3.5 client profile will compile with the nuget package. However the customer will need to manually enable TLS (server patching erc)

Mindscape.Raygun4Net.Common/DataAccess/RaygunWebClient.cs
```diff
@@ -0,0 +1,46 @@
...
+#if NET40 || NET45
+    private static readonly PropertyInfo SslProtocolsPropertyInfo =
+      typeof(HttpWebRequest).GetProperty("SslProtocols", BindingFlags.Instance | BindingFlags.NonPublic);
+#endif
+
+    public RaygunWebClient(SecurityProtocolType? sslProtocol = null)
+    {
+      SecurityProtocolType defaults = ServicePointManager.SecurityProtocol;
+#if NET40
+      defaults |= (SecurityProtocolType)3072; //TLS 1.2
```
Olwiba, line 23: How come we can't use the enum for NET40?

Olwiba (commented):
Just had a question around the use of enums for NET 4

Mindscape.Raygun4Net.Common/DataAccess/RaygunWebClient.cs
darcythomas: So an enum is kinda like a wrapper for an int  (in this case), with names fore certain values.

In the higher versions of .net we have the tls1.1-3 defined as names enums. 
However the places that read from this enum are only using the int (or more accurately byte) value.
(my guess is that the C# call makes an interop call to the windows (winkernal32.dll?) C++ code and passes the bits across. because C# enums are not a thing in C++.


What this means is as long as we pass along a valid byte value for the tls version we want to use it will get passed down to the OS and it will work.


If the network stack for .net 4-4.5 was written in managed code (C#) then this may not have worked.
In that respect we are lucky.

As to why I used the named enums when I could?  Well it doesn't make any function difference but if people were not familiar with what is going on it may help them and it kinda shows where the magic numbers come from.

darcythomas: TL;DR; Because the value is not a named value in the enum in that version of .net. But the int value works just fine (I made sure to test it)

Olwiba: gotcha, thanks for clearing that up 😃 I figured it would be something like that

buildSigned.ps1
xenolightning: 😆 - good catch!

Here I was thinking I could use this for other multi-targeting issues we had with psake!

darcythomas: FWIW adding `Framework "4.8"` at the top of a psake file will get  it to use VS v16 (2019)
 
and `dotnet msbuild` uses what ever is installed, so in my case v17 (2022)

darcythomas: I have reverted.  The actual problem was the `/m` flag which was breaking the build

darcythomas:
I have add to make some more chages to get it to package correctly. 
Commented out the windows RT and phone stuff from the build and package scripts.


It all builds and packages up nicely now.

Mindscape.Raygun4Net4.Intergration.Tests/TestIfWeCanSendToRaygun.cs
QuantumNightmare: not a problem :+1:

Mindscape.Raygun4Net.Common/DataAccess/RaygunWebClient.cs
```diff
@@ -0,0 +1,46 @@
...
+  public class RaygunWebClient : WebClient
+  {
+    private readonly SecurityProtocolType _sslProtocol;
+
+#if NET40 || NET45
+    private static readonly PropertyInfo SslProtocolsPropertyInfo =
+      typeof(HttpWebRequest).GetProperty("SslProtocols", BindingFlags.Instance | BindingFlags.NonPublic);
+#endif
+
+    public RaygunWebClient(SecurityProtocolType? sslProtocol = null)
```
QuantumNightmare, line 19: I didn't see if sslProtocol is ever passed into this constructor. If not, wondering what your plans for this are.

darcythomas: No I don't use it.

My thinking was this would be an 'escape hatch', if we needed to override the default behaviour

QuantumNightmare (dismissed):
Looking good to me - just added one more comment

Mindscape.Raygun4Net.Common/DataAccess/RaygunWebClient.cs
QuantumNightmare: Good to know 👌

xenolightning (dismissed):
Top work Darcy 😎

Olwiba (dismissed):
Ace ♠️

Mindscape.Raygun4Net4.ClientProfile.Tests/Setup/TraceChecker.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mindscape.Raygun4Net4.Integration.Tests.Setup
+{
+
+  [ExcludeFromCodeCoverage]
```
Olwiba, line 8: sorry if this is a stupid question:
how come we've excluded this from code coverage? does this have the possibility of introducing a tripping point in the future?

Olwiba (dismissed):
Changes from the last review look fine to me. Nice work adding extra tests 🔎

Mindscape.Raygun4Net4.ClientProfile.Tests/Setup/TraceChecker.cs
darcythomas: It's a good question!
This code is needed so we can see what Traces (think internal logs) the Raygun client is creating, at unit test time.


The `RaygunClient` effectively **only** uses the `WriteLine(string message)` method.
However the `TraceListener` base class is an abstract class, and we have to implement all of the abstract methods.

The unit tests never use the ` Write(string message)` so it always shows as uncovered (and an ugly red color, shocking!)

Since this is not critically important code I thought it better to exclude it from coverage rather than add 'noise'.

I guess I could have just put the `[ExcludeFromCodeCoverage]` attribute onto **just** the unused methods. I may do that next time.

Olwiba: thanks for the explanation!

Olwiba (approved):
Providing testing looked good, I'm happy! Nice work 🚀

PR #480, opened 2023-06-06
[ACT-64] Update ASP.NET Framework MVC and WebAPI READMEs

crystalqdm:
This PR is to update the ASP.NET WebAPI and MVC README documentation so that they are clear about being relevant for ASP.NET Framework, and where to go for ASP.NET Core docs


Hamish-taylor (approved):
Changes look good to me!

PR #487, opened 2023-08-15
Adjust projects from AllStandardProjects.sln so they can build in Rider

phillip-haydon:
Was unable to build the projects as they targeted old versions of .NET Framework, and debug builds required the signing key.

- Added Microsoft.NETFramework.ReferenceAssemblies.* for older versions of .NET Framework
- Bumped test projects to .NET 4.8
- Moved the signing key to not be required in DEBUG builds
- Made integration tests fail outright with an up front fail for the API Key


darcythomas (approved):
LGTM

PR #488, opened 2023-08-16
Make RaygunClient non blocking for SendInBackground

phillip-haydon:
- Renamed `Mindscape.Raygun4Net.AspNetCore2` to `Mindscape.Raygun4Net.AspNetCore` and bumped to .NET 6
- Made the SendInBackground not await
- Added test to verify it sends and isn't blocking


Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -275,86 +280,65 @@ public void Send(Exception exception, IList<string> tags)
...
     }
 
     /// <summary>
-    /// Asynchronously transmits a message to Raygun.
+    /// Transmits an exception to Raygun in the background without blocking.
     /// </summary>
     /// <param name="raygunMessage">The RaygunMessage to send. This needs its OccurredOn property
     /// set to a valid DateTime and as much of the Details property as is available.</param>
-    public Task SendInBackground(RaygunMessage raygunMessage)
+    public void SendInBackground(RaygunMessage raygunMessage)
```
darcythomas, line 330: This could be a breaking change. (if someone is awaiting this call.

You could maybe return a `Task.CompletedTask` (and still leave the un-awated send as you have  now)

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -275,86 +280,65 @@ public void Send(Exception exception, IList<string> tags)
...
-
-    /// <summary>
-    /// Asynchronously transmits an exception to Raygun.
-    /// </summary>
-    /// <param name="exception">The exception to deliver.</param>
-    /// <param name="tags">A list of strings associated with the message.</param>
-    public Task SendInBackground(Exception exception, IList<string> tags)
-    {
-      return SendInBackground(exception, tags, null, null);
+      SendAsync(exception, tags, userCustomData).GetAwaiter().GetResult();
```
darcythomas, line 283: This looks like code smell. But I think this is the exception to the rule.

Would it be worth wrapping in in a "try catch <s>log,/s> `Trace` swallow' " pattern?

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -275,86 +280,65 @@ public void Send(Exception exception, IList<string> tags)
...
     /// <param name="tags">A list of strings associated with the message.</param>
     /// <param name="userCustomData">A key-value collection of custom data that will be added to the payload.</param>
-    public Task SendInBackground(Exception exception, IList<string> tags, IDictionary userCustomData)
+    /// <param name="userInfo">Information about the user including the identity string.</param>
+    public virtual async Task SendAsync(Exception exception, IList<string> tags, IDictionary userCustomData, RaygunIdentifierMessage userInfo = null)
     {
-      return SendInBackground(exception, tags, userCustomData, null);
+      if (CanSend(exception))
+      {
+        await StripAndSend(exception, tags, userCustomData, userInfo).ConfigureAwait(false);
```
darcythomas, line 297: Is  `.ConfigureAwait(false)` needed here?

https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-await-over-continuewith

Mindscape.Raygun4Net.NetCore.Tests/Mindscape.Raygun4Net.NetCore.Tests.csproj
```diff
@@ -2,19 +2,21 @@
   <PropertyGroup>
     <!--Out of support versions which still compiles/works @ May 2022-->
     <!--You may need to install extra SDK/Targeting packs to run these tests -->
-    <TargetFrameworks>netcoreapp3.0;netcoreapp2.2;netcoreapp2.1;netcoreapp2.0;netcoreapp1.1;netcoreapp1.0;net461;</TargetFrameworks>
+    <TargetFrameworks>net5.0;netcoreapp3.1;netcoreapp3.0;netcoreapp2.2;netcoreapp2.1;netcoreapp2.0;netcoreapp1.1;netcoreapp1.0;net461;</TargetFrameworks>
    
     <!-- Versions which are supported by Microsoft @ May 2022-->
-    <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;net48;net472;net471;net47;net462</TargetFrameworks>
+    <TargetFrameworks>net6.0;net48;net472;net471;net47;net462</TargetFrameworks>
```
darcythomas, line 8: Not for today. But it would be good to add .net7 and .net8 (preview) here. May need to do this in docker/ github actions to make less overhead to get running

Mindscape.Raygun4Net.NetCore.Tests/RaygunClientIntegrationTests.cs
```diff
@@ -0,0 +1,74 @@
...
+      // 50 * 300 = 15000ms
+      // If the requests were blocking it would take minimum 15 seconds at minimum 300ms per request to send each one
+      // So we can assume that the requests are being sent in parallel, this should be a lot less than 15 seconds
+      Assert.That(elapsed, Is.LessThan(500));
+
+      // Delay 2 seconds to give it time to send all the messages
+      await Task.Delay(2000);
+
+      // Verify that the request was sent 50 times
+      await mockHttp.VerifyAsync(match => match.Method(HttpMethod.Post)
```
darcythomas, line 70: Nice!

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
phillip-haydon: Good point! I'll update this.

Mindscape.Raygun4Net.NetCore.Tests/Mindscape.Raygun4Net.NetCore.Tests.csproj
darcythomas: Also dead comment above

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
phillip-haydon: So the only real change here is `.Wait` will cause the exception to be wrapped in an `AggregateException`, calling `.GetAwaiter().GetResult()` will throw the original exception not the wrapped one.

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
phillip-haydon: `StripAndSend` already does this so it shouldn't be needed. We are just passing the Task back and not doing anything in that awaiting anything else.

Mindscape.Raygun4Net.NetCore.Tests/Mindscape.Raygun4Net.NetCore.Tests.csproj
phillip-haydon: I concur.

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -327,9 +327,10 @@ public virtual async Task SendAsync(Exception exception, IList<string> tags, IDi
     /// </summary>
     /// <param name="raygunMessage">The RaygunMessage to send. This needs its OccurredOn property
     /// set to a valid DateTime and as much of the Details property as is available.</param>
-    public void SendInBackground(RaygunMessage raygunMessage)
+    public Task SendInBackground(RaygunMessage raygunMessage)
```
darcythomas, line 330:
```suggestion
        public Task SendInBackground(RaygunMessage raygunMessage)
    {
       return   SendInBackgroundButReturnCompletedTask(raygunMessage);
    }
    
    ///<summary>For backwards compatibility we need to return a Task. 
    /// But we DO NOT want to block/await on SendInBackground(). 
    /// Which is why we _ discard the Task.Run() and return a Task.CompletedTask
    ///</summary>
    private Task SendInBackgroundButReturnCompletedTask(RaygunMessage raygunMessage)
    {    
       _ =   Task.Run(() => Send(raygunMessage));
      return Task.CompletedTask;
    }
    
```

What we are doing here with Tasks looks 'weird'. 
But we have our reasons. 
Comment rot is a thing. 

So suggesting pulling this into a private method which is a bit more explicit

darcythomas (dismissed):
One optional suggestion.

LGTM

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
darcythomas: Out of scope for this PR, but I think we should wrap all of our public APIs like this one in a `try catch swallow` pattern

Because not sending an error log is a level 1 sin
But crashing the app when trying to send a log is a level 9

QuantumNightmare: Except if RaygunSettings.ThrowOnError is set to true, then the error should be thrown. (This is a mechanism used to help debug issues with the provider)

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
phillip-haydon: I added comment but i applied it as a normal comment not an xml comment so we don't need to extract another method.

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
phillip-haydon: Added try/catch and throw if `RaygunSettings.ThrowOnError` is true.

darcythomas (approved):
LGTM

PR #495, opened 2023-09-12
Adjust Nuspec to fix wrong dll being packaged

phillip-haydon:
When migrating all the nuspecs i packaged the wrong one. Mindscape.Raygun4Net.Core is actually built as Mindscape.Raygun4Net so 2 packages were being bundled with the wrong dll. This has been fixed and deployed to Nuget already as 7.0.1

### DIFF
--- a/GlobalAssemblyInfo.cs
+++ b/GlobalAssemblyInfo.cs
@@ -3,5 +3,5 @@
 
 using System.Reflection;
 
-[assembly: AssemblyFileVersion("")]
-[assembly: AssemblyInformationalVersion("")]
+[assembly: AssemblyFileVersion("7.0.1")]
+[assembly: AssemblyInformationalVersion("7.0.1")]
--- a/GlobalAssemblyInfo.cs
+++ b/GlobalAssemblyInfo.cs
@@ -3,5 +3,5 @@
 
 using System.Reflection;
 
-[assembly: AssemblyFileVersion("7.0.1")]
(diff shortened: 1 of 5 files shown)


GlobalAssemblyInfo.cs
```diff
@@ -3,5 +3,5 @@
 
 using System.Reflection;
 
-[assembly: AssemblyFileVersion("")]
-[assembly: AssemblyInformationalVersion("")]
+[assembly: AssemblyFileVersion("7.0.0")]
```
darcythomas, line 6:
```suggestion
[assembly: AssemblyFileVersion("7.0.1")]
```

darcythomas: Should this be 7.0.1?

GlobalAssemblyInfo.cs
```diff
@@ -3,5 +3,5 @@
...
+[assembly: AssemblyInformationalVersion("7.0.0")]
```
darcythomas, line 7:
```suggestion
[assembly: AssemblyInformationalVersion("7.0.1")]
```

darcythomas (approved):
Looks good.

Just check on the version number in the GlobalAsssemblyInfo.cs

GlobalAssemblyInfo.cs
phillip-haydon: I just left the value in from testing locally, these are set in TeamCity during the build process.

PR #496, opened 2023-09-27
Updated environment information

phillip-haydon:
- Fixed Environment Information Memory
- Added cache for Environment Information to reduce overhead
- Added support for Environment Information on Windows/Linux/MacOS
- Changed CPU / OS information to be more accurate / understandable

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Common/Builders/RaygunEnvironmentMessageBuilder.cs
+++ b/Mindscape.Raygun4Net.NetCore.Common/Builders/RaygunEnvironmentMessageBuilder.cs
@@ -1,41 +1,75 @@
-﻿
-using System;
-using System.IO;
+﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
-using System.Linq;
 using System.Runtime.InteropServices;
+using System.Threading;
+using Mindscape.Raygun4Net.EnvironmentProviders;
 
 namespace Mindscape.Raygun4Net
 {
   public class RaygunEnvironmentMessageBuilder
   {
-    public static RaygunEnvironmentMessage Build(RaygunSettingsBase settings)
-    {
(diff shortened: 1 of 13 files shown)


Mindscape.Raygun4Net.NetCore.Common/Builders/RaygunEnvironmentMessageBuilder.cs
```diff
@@ -1,41 +1,78 @@
...
+using Mindscape.Raygun4Net.EnvironmentProviders;
 
 namespace Mindscape.Raygun4Net
 {
   public class RaygunEnvironmentMessageBuilder
   {
-    public static RaygunEnvironmentMessage Build(RaygunSettingsBase settings)
-    {
-      RaygunEnvironmentMessage message = new RaygunEnvironmentMessage();
+    private static readonly RaygunEnvironmentMessage CachedMessage = new RaygunEnvironmentMessage();
```
xenolightning, line 12: The only thing I can think of here - that by caching the message - we do open ourselves up to that private field being modified outside the `Build()` method.

Because it's the same reference returned I could be a bad person do something like 

```
RaygunEnvironmentMessageBuilder.Build(_).OSVersion = "this.phil"
```

And that would cause the OSVersion to be modified for all following reports..? Maybe there's a way to make this immutable?

image

xenolightning (changes requested):

Mindscape.Raygun4Net.NetCore.Common/Builders/RaygunEnvironmentMessageBuilder.cs
phillip-haydon: Now returning a new object to avoid the original being modified.

xenolightning: I like it 😎

xenolightning (approved):

PR #499, opened 2023-11-14
Add the missing AssemblyVersion to AssemblyInfo

phillip-haydon:
To fix dependent projects referencing the wrong version this adds the missing `AssemblyVersion` used by StrongName.


CmdrKeen (approved):
Looks good to me. 

As discussed in Slack, the version field updated to go back from 7.1.0 -> 7.0.0 doesn't matter, it's designed to be updated from the build server so the value going backwards won't be an issue.

PR #500, opened 2023-11-14
Add missing change-log for v8.0.0

phillip-haydon:
Missed the change log for v8.0.0


MattByers (approved):
That was a tough one 😅

PR #501, opened 2023-11-19
[CS-55] Dont block sending

darcythomas:
DRAFT Looking for comments/feedback

Context: https://raygun.atlassian.net/browse/CS-55

Sending to RG in some cases, seems to be blocking (the UI thread in RG4MAUI)

Sorry: there is some code weeding noise in there too.

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Common/FireAndForget.cs
+++ b/Mindscape.Raygun4Net.NetCore.Common/FireAndForget.cs
@@ -1,0 +1,123 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Mindscape.Raygun4Net
+{
+  internal class Fire
+  {
+
+    /// <summary>
+    /// If you need to see the details of an exception thrown please, add a logging action here.
+    /// </summary>
+    internal static Action<string> DebugLog { get; set; } = message => { /* NOOP by default */};
+
+    /// <summary>
+    /// If you want to return as soon as possible (and not block or await)
+    /// Pass in an action into this.
+    /// </summary>
+    /// <remarks>
+    /// This method will swallow *all* exceptions thrown.
+    /// I am not your therapist. Please deal with your own exceptions</remarks>
+    /// <param name="action"></param>
+    /// <param name="memberName"></param>
+    /// <param name="sourceLineNumber"></param>
+    internal static void AndForget(
+      Action action,
+      [System.Runtime.CompilerServices.CallerMemberName]
+      string memberName = "",
+      [System.Runtime.CompilerServices.CallerLineNumber]
+      int sourceLineNumber = 0)
+    {
+      Task ActionAsTask()
+      {
+        action();
+        return Task.CompletedTask;
+      }
+
+      AndForget(ActionAsTask,
+        memberName,
+        // ReSharper disable once ExplicitCallerInfoArgument
+        sourceLineNumber
+        );
+    }
+
+    /// <summary>
+    /// If you want to return as soon as possible (and not block or await)
+    /// Pass in an action into this.
+    /// </summary>
+    /// <remarks>
+    /// This method will swallow *all* exceptions thrown.
+    /// I am not your therapist. Please deal with your own exceptions</remarks>
+    /// <param name="task"></param>
+    /// <param name="memberName"></param>
+    /// <param name="sourceLineNumber"></param>
+    internal static void AndForget(
+      Task task,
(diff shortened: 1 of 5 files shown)


Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -436,6 +437,24 @@ protected IEnumerable<Exception> StripWrapperExceptions(Exception exception)
     /// <param name="raygunMessage">The RaygunMessage to send. This needs its OccurredOn property
     /// set to a valid DateTime and as much of the Details property as is available.</param>
     public async Task Send(RaygunMessage raygunMessage)
+    {
```
darcythomas, line 440: Magic here

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -436,6 +437,24 @@ protected IEnumerable<Exception> StripWrapperExceptions(Exception exception)
...
+      //ONLY when debugging do we want to let an error bubble up!
+      if (Settings.ThrowOnError)
+      {
+        await SendInternal(raygunMessage);
```
xenolightning, line 444: We should `ConfigureAwait(false)` this? To avoid a deadlock if say a UI thread raises this directly when throwing is on?

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -10,18 +10,19 @@
...
   {
+    protected internal const string SentKey = "AlreadySentByRaygun";
+
     private static HttpClient Client = new HttpClient();
+    private bool _handlingRecursiveErrorSending;
+    private bool _handlingRecursiveGrouping;
 
     private readonly string _apiKey;
     private readonly List<Type> _wrapperExceptions = new List<Type>();
+    protected readonly RaygunSettingsBase Settings;
```
xenolightning, line 24: We might need to bump a major version if we change this, it's technically a public api change?

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
xenolightning: It might not matter because the actual async operation is also configured, but could be safe guard

Mindscape.Raygun4Net.NetCore.Common/FireAndForget.cs
```diff
@@ -0,0 +1,123 @@
...
+
+namespace Mindscape.Raygun4Net
+{
+  internal class Fire
+  {
+
+    /// <summary>
+    /// If you need to see the details of an exception thrown please, add a logging action here.
+    /// </summary>
+    internal static Action<string> DebugLog { get; set; } = message => { /* NOOP by default */};
```
xenolightning, line 12: Is this property meant to be something that is user configurable? I note that it's not settable in the public api surface, and we don't provide a way to override it?

xenolightning (changes requested):
Just a couple of comments/questions

Mindscape.Raygun4Net.NetCore.Common/FireAndForget.cs
darcythomas: Currently we have a if (_settings.ThrowOnError) for debugging.

However that is pretty heavy handed and would not work where this is used because Fire.AndForget explicitly swallows any errors.

I have ideas for making a  public debug log (even if we tag it  that we make no guaranteed to support it etc)
Similar to what we have in raygun4js.


I would also like to have some on onboarding setup debug helper (maintained API) (e.g., you apikey is empty, invalid, there is no network connection etc)


That is all future direction stuff.
This is just enough for us to have a way of debugging

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
darcythomas: Reverted

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
darcythomas: Should not matter, but since we don't want to mess up someone else's app. Better safe than sorry.

darcythomas:
Closing for now

PR #503, opened 2023-12-27
Mindscape.Raygun4Net.NetCore.Common - Remove unnecessary nuget dependencies

snakefoot:
Removed:
- System.Runtime.Serialization.Primitives ver. 4.3.0
- System.Management ver. 4.7.0

### DIFF
--- a/CHANGE-LOG.md
+++ b/CHANGE-LOG.md
@@ -1,5 +1,9 @@
 # Full Change Log for Raygun4Net.* packages
 
+### v8.0.2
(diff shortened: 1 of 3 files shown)


xenolightning:
Hi @snakefoot 

Thanks for the PR, I think this change is good.

We did notice these extra dependencies late last year, and we have another branch cleaning up unhandled exceptions too.

To avoid any version conflicts and testing, I'll make this change in that branch and I'll tag you on that PR FYI (when it's ready for review) 

I'll leave this open for now though

PR #504, opened 2024-01-12
Cross Platform global exception handling

xenolightning:
Add low level global hooks for unhandled events across multiple platforms.

This should solve the issues with not catching unhandled exceptions.

There is also no workload requirements due to using reflection everywhere.


Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -101,11 +104,8 @@ public RaygunClientBase(RaygunSettingsBase settings)
         ApplicationVersion = settings.ApplicationVersion;
       }
 
-#if NETSTANDARD2_0_OR_GREATER
-      if (_settings.CatchUnhandledExceptions)
-      {
-         CatchUnhandledExceptions = _settings.CatchUnhandledExceptions;
-      }
+#if UNHANDLED_EXCEPTIONS_SUPPORT
+      UnhandledExceptionBridge.OnUnhandledException += OnApplicationUnhandledException;
```
xenolightning, line 108: A big change here is always subscribing to the unhandled exceptions. So the callbacks are fired each and everytime.

They are then filtered out before sending.

This may impact performance - but since it's on unhandled exceptions, when the app is misbehaving or about to crash it should be fine???

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.0",
+    "version": "7.0.0",
```
xenolightning, line 3: this can be put back to 6.0.0 - was done for maui testing

Mindscape.Raygun4Net.WebApi/Mindscape.Raygun4Net.WebApi.csproj
```diff
@@ -66,6 +66,7 @@
     </Compile>
     <Compile Include="..\Mindscape.Raygun4Net.Common\DataAccess\RaygunWebClient.cs">
       <Link>RaygunWebClient.cs</Link>
+      <SubType>Component</SubType>
```
xenolightning, line 69: Remove

Mindscape.Raygun4Net.NetCore.Common/Messages/RaygunClientMessage.cs
```diff
@@ -1,12 +1,26 @@
...
+    private static readonly string ClientVersion;
+
+    static RaygunClientMessage()
+    {
+      var informationalVersion = Assembly
+                          .GetExecutingAssembly()
+                          .GetCustomAttribute<AssemblyInformationalVersionAttribute>()
+                          ?.InformationalVersion;
+
+      ClientVersion = informationalVersion ?? "8.1.0";
```
phillip-haydon, line 16: Is there any reason why we would never get a value here other than testing? Worried this might be something we need to manually update all the time.

phillip-haydon: (the fix its for is good tho)

Mindscape.Raygun4Net.NetCore.Common/Platforms/AndroidPlatform.cs
```diff
@@ -0,0 +1,62 @@
...
+  {
+    private static readonly Assembly AndroidAssembly =
+      AssemblyHelpers.FindAssembly("Mono.Android", AssemblyHelpers.HexStringToByteArray("84e04ff9cfb79065"));
+
+    public static bool TryAttachExceptionHandlers()
+    {
+      try
+      {
+        if (AndroidAssembly is null)
+          return false;
```
phillip-haydon, line 17: Missing braces!

Mindscape.Raygun4Net.NetCore.Common/Platforms/ApplePlatform.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    public static bool TryAttachExceptionHandlers()
+    {
+      try
+      {
+        // One or the other, the types and names are the same across both
+        var activeAssembly = IOSAssembly ?? MacCatalystAssembly;
+
+        if (activeAssembly is null)
+          return false;
```
phillip-haydon, line 25: Missing braces!

Mindscape.Raygun4Net.NetCore.Common/Platforms/WindowsPlatform.cs
```diff
@@ -0,0 +1,77 @@
...
+      AssemblyHelpers.FindAssembly("Microsoft.WinUI", AssemblyHelpers.HexStringToByteArray("de31ebe4ad15742b"));
+
+    private static Exception _lastFirstChanceException;
+
+    public static bool TryAttachExceptionHandlers()
+    {
+      try
+      {
+        if (WinUIAssembly is null)
+          return false;
```
phillip-haydon, line 19: 👀

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -57,31 +62,29 @@ public abstract class RaygunClientBase
...
-        {
-          Send(e.ExceptionObject as Exception, new[] { "UnhandledException" });
-        }
       }
     }
 
+    private void OnApplicationUnhandledException(Exception exception, bool isTerminating)
+    {
+      if (!CatchUnhandledExceptions)
+        return;
```
phillip-haydon, line 83: 👀

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -48,7 +53,7 @@ public abstract class RaygunClientBase
     /// </summary>
     public string ApplicationVersion { get; set; }
 
-#if NETSTANDARD2_0_OR_GREATER
+#if UNHANDLED_EXCEPTIONS_SUPPORT
```
phillip-haydon, line 56: Is this still required? If we are doing reflection to wire things up can we not just set it to wire up always even if by default its an empty handler? Wondering if it helps clean up the code and/or prevent the requirement of precomplication flags.

phillip-haydon (changes requested):
Needs more braces :D

Mindscape.Raygun4Net.NetCore.Common/Platforms/AndroidPlatform.cs
xenolightning: 👿

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
xenolightning: Have removed pre-processor def's

The targets always include this particular flag anyway

Mindscape.Raygun4Net.NetCore.Common/Messages/RaygunClientMessage.cs
xenolightning: It's unlikely - I set it as a fallback so we have _something_. In the unlikely scenario we are not able to use reflection.

The last value set was 6.5.x so it wasn't updated as it was.

If we update it, cool, if we don't, cool.

Mindscape.Raygun4Net.NetCore.Common/Platforms/AndroidPlatform.cs
phillip-haydon: 🤤

phillip-haydon (approved):
:shipit:

phillip-haydon (dismissed):
:shipit:

phillip-haydon (approved):
:shipit: NOW! No more additions please :D

xenolightning:
Changes to be rolled into a later 8.2.0 nuget release with support for httpclient and timeouts

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
Obi-Dann: In the context of AspNetCore, there will be multiple instances of RaygunClient created for each HttpContext. Every single one of these instances will subscribe to OnUnhandledException.
None of those instances will be garbage collected because currently RaygunClientBase is not disposable and never unsubscribes from OnUnhandledException.
I suspect in those scenarios, an unhandled rejection will be send as many times as many clients were ever created.
What is the expected lifetime of RaygunClient? This change assumes it's a singleton, but it's not used like a singleton in AspNetCore.
It's not fine

PR #506, opened 2024-01-24
Optimise resource consumption & drop support for EOL NETFX

xenolightning:
Changes Summary:
* Add ThrottledBackgroundMessageProcessor to prevent thread and SNAT exhaustion
* Remove targeting unsupported NETFX versions. Only target 4.6.2+
* Convert projects to SDK style
* Remove old NETFX codebase - and pull into new projects
* Remove Client Profile targets
*


Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
```diff
@@ -0,0 +1,113 @@
...
+      _messageQueue = new BlockingCollection<RaygunMessage>(maxQueueSize);
+      _cancelProcessingSource = new CancellationTokenSource();
+      _workerTasks = new List<Task>();
+    }
+
+    public bool Enqueue(RaygunMessage message)
+    {
+      var itemAdded = _messageQueue.TryAdd(message);
+
+      EnsureWorkers();
```
xenolightning, line 37: Instead of having another thread, or a periodic timer, just ensure that there are workers available when queuing a message

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
```diff
@@ -0,0 +1,113 @@
...
+
+      return itemAdded;
+    }
+
+    private void EnsureWorkers()
+    {
+      lock (_workerTaskMutex)
```
xenolightning, line 44: This lock could be moved below, but the active workers could change in this scope, so I wrapped everything in the lock for now

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+          // Minus one from the active worker count
+          Interlocked.Decrement(ref _activeWorkers);
+        });
+
+      Interlocked.Increment(ref _activeWorkers);
+      return workerTask;
+    }
+
+    private static async Task RaygunMessageWorker(BlockingCollection<RaygunMessage> messageQueue, Func<RaygunMessage, CancellationToken, Task> callback,
```
xenolightning, line 93: Decided to make this a static "pure" function, so it doesn't leak state into the wider class.

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+      var workerTask = Task.Run(async () =>
+        {
+          // Run the message processor loop
+          await RaygunMessageWorker(_messageQueue, _processCallback, _cancelProcessingSource.Token);
+        })
+        .ContinueWith(x =>
+        {
+          // Minus one from the active worker count
+          Interlocked.Decrement(ref _activeWorkers);
```
xenolightning, line 86: Old school syntax, because it suits this use case a little better.

Mindscape.Raygun4Net.NetCore.Common/RaygunSettings.cs
```diff
@@ -26,5 +34,15 @@ public RaygunSettingsBase()
...
+
+    public ushort BackgroundMessageQueueMax { get; } = 10_000;
+
+    /// <summary>
+    /// Controls the number of background threads used to process the raygun message queue
+    /// </summary>
+    /// <remarks>
+    /// Defaults to Environment.ProcessorCount * 2 &gt;= 8 ? 8 : Environment.ProcessorCount * 2
+    /// </remarks>
+    public int BackgroundMessageWorkerCount { get; set; } = Environment.ProcessorCount * 2 >= 8 ? 8 : Environment.ProcessorCount * 2;
```
xenolightning, line 46: This value is TBD.

Noting that on my development machine I have 20 CPU Cores, and 40 background threads is a bit overkill. So I capped it at 8 by default.

Mindscape.Raygun4Net.NetCore.Tests/RaygunClientIntegrationTests.cs
xenolightning, line 1: These tests are janky, and I'll fix them up, and also unit the actual class

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -335,23 +336,22 @@ public virtual async Task SendAsync(Exception exception, IList<string> tags, IDi
...
-        // So we just fire and forget and discard the task.
-        // If we await this Task it will cause SendInBackground to be blocking.
-        _ = Task.Run(async () =>
+        var exceptions = StripWrapperExceptions(exception);
+        foreach (var ex in exceptions)
         {
-          await StripAndSend(exception, tags, userCustomData, userInfo);
-        });
-
+          var message = await BuildMessage(ex, tags, userCustomData, userInfo).ConfigureAwait(false);
```
xenolightning, line 346: We can actually await things now.

Since the enqueuing is done synchronously, and the processing is done by a background thread, it's totally fine to do this.

Note, I moved the implementation from stripandsend, to what should actually be done

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
xenolightning: I have replaced with a monitor to better control entry & exit

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
```diff
@@ -0,0 +1,116 @@
...
+    private readonly BlockingCollection<RaygunMessage> _messageQueue;
+    private readonly List<Task> _workerTasks;
+    private readonly CancellationTokenSource _cancelProcessingSource;
+    private readonly Func<RaygunMessage, CancellationToken, Task> _processCallback;
+    private readonly int _maxWorkerTasks;
+    private readonly object _workerTaskMutex = new object();
+
+    private volatile bool _isDisposing;
+
+    public ThrottledBackgroundMessageProcessor(int maxQueueSize, int maxWorkerTasks, Func<RaygunMessage, CancellationToken, Task> onProcessMessageFunc)
```
xenolightning, line 23: @PanosNB - would be good to get you to review this class.

It's fairly concise, for NETFX I'll create a copy of this and remove the async/await syntax

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -22,7 +23,7 @@ public abstract class RaygunClientBase
       // The default timeout is 100 seconds for the HttpClient, 
       Timeout = TimeSpan.FromSeconds(30)
     };
```
Obi-Dann, line 25: One other thing that can prevent SNAT and thread exhausting is setting `MaxConnectionsPerServer` to something less than an unlimited number by default. It can be achieved by passing HttpClientHandler

```c#
new HttpClient(new HttpClientHandler
  {
      MaxConnectionsPerServer = 50
  }) {
  Timeout = TimeSpan.FromSeconds(30)
  };
```

xenolightning: We did initially look at this as a solution. 

This issue with this is that is silently drops errors from being sent, and in our testing it doesn't always solve the SNAT issue 😿

phillip-haydon: We actually tested this initially and found it didn't have any major contribution of reducing the number of ports taken, for example when set to 10, i was still hitting around ~2000 ports being taken in Azure and of 50k requests, only around ~4k would succeed, while leaving it defaulted it tried to use around ~3000 and ~49k requests succeeded.

Mindscape.Raygun4Net.Core/RaygunSettings.cs
```diff
@@ -22,6 +23,36 @@ public static RaygunSettings Settings
...
+      if (!string.IsNullOrEmpty(messageQueueMaxValue) && int.TryParse(messageQueueMaxValue, out var maxQueueSize))
+      {
+        BackgroundMessageQueueMax = maxQueueSize;
+      }
+    }
+    
+    /// <summary>
+    /// The maximum queue size for background exceptions
+    /// </summary>
+    public int BackgroundMessageQueueMax { get; } = ushort.MaxValue;
```
PanosNB, line 41: No environment configuration possible for this one?

Mindscape.Raygun4Net.NetCore.Common/RaygunSettings.cs
```diff
@@ -5,14 +5,22 @@ namespace Mindscape.Raygun4Net
...
   {
     internal const string DefaultApiEndPoint = "https://api.raygun.com/entries";
+    private const string RaygunMessageQueueMaxVariable = "RAYGUN_MESSAGE_QUEUE_MAX";
 
     public RaygunSettingsBase()
     {
       ApiEndpoint = new Uri(DefaultApiEndPoint);
+      
+      // See if there's an overload defined in an environment variable, and set it accordingly
+      var messageQueueMaxValue = Environment.GetEnvironmentVariable(RaygunMessageQueueMaxVariable);
```
PanosNB, line 15: This continues from my comment above. So we check for environment initialization separately?

Mindscape.Raygun4Net.NetCore.Common/RaygunSettings.cs
PanosNB: There is initialization of these values in multiple areas? Here, it's missing the environment override. Can we do it once and import the code to all

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
PanosNB: Why is this running regardless if TryAdd succeeded or not (due to max queue length crossed)?

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
```diff
@@ -0,0 +1,138 @@
...
+      try
+      {
+        // Remove dead/faulted/finished tasks
+        _workerTasks.RemoveAll(x => x.IsCompleted);
+
+        var numberOfWorkersToStart = _maxWorkerTasks - _workerTasks.Count;
+
+        for (var i = 0; i < numberOfWorkersToStart; i++)
+        {
+          _workerTasks.Add(CreateWorkerTask());
```
PanosNB, line 73: I see, so the thread pool starts empty and then this call on the first SendInBg will fill up all of them (and the first will pick up the task added in ln 49 above). Plus, on every SendInBg call, EnsureWorkers acts as a health-checking mechanism

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
```diff
@@ -0,0 +1,138 @@
...
+    private Task CreateWorkerTask()
+    {
+      var workerTask = Task.Run(async () =>
+      {
+        // Run the message processor loop
+        await RaygunMessageWorker(_messageQueue, _processCallback, _cancelProcessingSource.Token);
+      });
+
+      // When a worker finishes ensure that a new one is is created if required
+      workerTask.ContinueWith(x => { EnsureWorkers(); });
```
PanosNB, line 108: Should this execute in a finally block and all the above in try?

PanosNB: Ignore, I now see this is still running on the customer thread and you are declaring a follow-up action, to invoke EnsureWorkers

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
```diff
@@ -0,0 +1,138 @@
...
+      }
+      catch (Exception cancelledEx) when (cancelledEx is ThreadAbortException ||
+                                          cancelledEx is OperationCanceledException ||
+                                          cancelledEx is TaskCanceledException)
+      {
+        // Cancellation was requested, so it's fine
+      }
+      catch (Exception ex)
+      {
+        Debug.WriteLine("Exception in queue worker {0}: {1}", Task.CurrentId, ex);
```
PanosNB, line 134: Hopefully, the passed tasks have their own try-catch logic, so something application-specific, like network errors, do not bubble up here

Mindscape.Raygun4Net.WebApi/RaygunWebApiClient.cs
```diff
@@ -539,7 +553,15 @@ public void SendInBackground(Exception exception, IList<string> tags, IDictionar
     /// set to a valid DateTime and as much of the Details property as is available.</param>
     public void SendInBackground(RaygunMessage raygunMessage)
     {
-      ThreadPool.QueueUserWorkItem(c => Send(raygunMessage));
+      SendInBackground(() => raygunMessage);
+    }
+    
+    public void SendInBackground(Func<RaygunMessage> raygunMessage)
```
PanosNB, line 559: I don't wanna break the spec but in the future, we should consider making this a boolean and make it return the TryAdd result

Mindscape.Raygun4Net4/ThrottledBackgroundMessageProcessor.cs
```diff
@@ -0,0 +1,130 @@
+#nullable enable
```
PanosNB, line 1: Again this file. I see that certain things must be duplicated :-(

PanosNB (approved):
Great work @xenolightning ! I am approving this as it looks good; just a few Qs for you

Mindscape.Raygun4Net.Core/RaygunSettings.cs
xenolightning: For this particular value, we want it to be configurable using the standard settings approach.

For the RAYGUN_MESSAGE_QUEUE_MAX variable - we wanted to give customers the ability to set the value, but it should be an exceptional case, and should not be configured using the settings object

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
xenolightning: Try add might return false - when the queue is full.

In the case that the queue is full, and there are no workers, we still want to "ensure workers" - so the queue does eventually get drained

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
xenolightning: Correct yep!

This is to optimise the usage of the threads. We shouldn't allocate threads if there's no work to do, so only create the workers on the first message to send.

From there, instead of managing a daemon worker, or a timer (which could fault or crash) we just queue on every message enqueued. This ensures we do the work to ensure there are workers to process something, when we have something to process.

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
xenolightning: Yep, spot on.

This is a continuation that happens after the worker task has completed/faulted. It allows us to ensure that there are workers to process messages if for whatever reason they fault

Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
xenolightning: Yeah - so we mainly call `Send()` here - which does have it's own internal exception handling. This is for the odd case that something unexpected happens, and we still have a trace log of it, if debug logging is enabled

Mindscape.Raygun4Net4/ThrottledBackgroundMessageProcessor.cs
xenolightning: Yeah sadly, the code here is ever so slightly different. There's no async await support in the NETFX code

Mindscape.Raygun4Net.WebApi/RaygunWebApiClient.cs
xenolightning: We could make this private, it's not in the public surface yet.

Alternatively we can make this particular overload return a boolean.

I followed the existing convention for the structure here, we could revise the api surface in a follow up version

Mindscape.Raygun4Net.NetCore.Common/RaygunSettings.cs
xenolightning: So these values are the default values / initial values. We could initialize both in the constructor.

This is merely syntactical sugar for ctor initialized properties without needing to be explicit

Mindscape.Raygun4Net.NetCore.Common/RaygunSettings.cs
xenolightning: Yeah checking this code inline is a bit more difficult because of the try parse, so I put it in the ctor.

We could move all defaults into the ctor  if you think

Mindscape.Raygun4Net4/Breadcrumbs/RaygunBreadcrumbs.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
```
phillip-haydon, line 5: 👀 we should do a cleanup of usings in next pr.

Mindscape.Raygun4Net.Tests/RaygunBreadcrumbsTests.cs
```diff
@@ -27,7 +28,7 @@ public void TearDown()
 
     [Test]
     public void Set_ClassName_MethodName_And_LineNumber_Automatically_If_Configured()
-    {
+    {
```
phillip-haydon, line 31: 👎

Mindscape.Raygun4Net.Tests/Model/FakeRaygunClient.cs
```diff
@@ -22,12 +19,12 @@ public RaygunMessage ExposeBuildMessage(Exception exception, [Optional] IList<st
...
-
-    public WebClient ExposeCreateWebClient()
+    
+    public IEnumerable<Exception> ExposeStripWrapperExceptions(Exception exception)
     {
-      return CreateWebClient();
+      return StripWrapperExceptions(exception);
     }
-
+
```
phillip-haydon, line 27: Whitespace giving me anxiety...

phillip-haydon (approved):
Looking good.

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -35,6 +36,7 @@ public abstract class RaygunClientBase
     private bool _handlingRecursiveGrouping;
 
     protected readonly RaygunSettingsBase _settings;
+    private readonly ThrottledBackgroundMessageProcessor _backgroundMessageProcessor;
```
Obi-Dann, line 39: Hmm, RaygunAspNetClient is still created per request, it means that users using RaygunAspNet won't issues with SNAT ports fixed because there will be multiple instances of the Clients?
Also, it's Disposable but never Disposed?

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -335,23 +336,21 @@ public virtual async Task SendAsync(Exception exception, IList<string> tags, IDi
...
-        // So we just fire and forget and discard the task.
-        // If we await this Task it will cause SendInBackground to be blocking.
-        _ = Task.Run(async () =>
+        var exceptions = StripWrapperExceptions(exception);
+        foreach (var ex in exceptions)
         {
-          await StripAndSend(exception, tags, userCustomData, userInfo);
-        });
-
+          if (!_backgroundMessageProcessor.Enqueue(async () => await BuildMessage(ex, tags, userCustomData, userInfo)))
```
Obi-Dann, line 346: This can result in an issue depending on when BuildMessage runs. In AspNetCore, BuildMessage attempts to read data from HttpContext. If `BuildMessage` is executed after HttpContext is disposed, it will either fail or won't get necessary data.
Shouldn't `BuildMessage` be called before enqueue and `Enqueue` should receive the built message rather than a `Func<RaygunMessage>`?

Mindscape.Raygun4Net.WebApi/RaygunWebApiClient.cs
```diff
@@ -620,11 +642,19 @@ private void Send(string message)
...
       {
         Send(BuildMessage(e, tags, userCustomData, currentTime));
       }
     }
+    
+    private void StripAndSendInBackground(Exception exception, IList<string> tags, IDictionary userCustomData, DateTime? currentTime)
+    {
+      foreach (var e in StripWrapperExceptions(exception))
+      {
+        SendInBackground(() => BuildMessage(e, tags, userCustomData, currentTime));
```
Obi-Dann, line 655: Potentially similar issue here if `BuildMessage` relies on resources fron the current thread/context
https://github.com/MindscapeHQ/raygun4net/pull/506/files#r1483673952

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
xenolightning: The HttpContext is accessed and the data captured before `BuildMessage` is called, so there shouldn't be a chance of accessing a disposed context.

It does this only in the AspNetCore specific version of the client.

There's also issues with using ThreadLocal vs AsyncLocal - as raised in another GH issue - which will also cause some problems capturing the correct information.

Using the middleware, the clients that are created, never actually use this overload, and aren't using the background message queue either - which is something we should also fix.

If you don't mind bearing with us as we uncover some of these interesting code nuggets from the past, and find resolutions to them.

Cheers,
Sean

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
xenolightning: Using the middleware, you are correct, because the Client only controls it's own limitations, creating other clients gets around the throttling. This is a larger design choice we want to resolve.

Ideally the middleware shouldn't be creating a new client for each request - but it adds complexity.

Alternatively, the message queue could be a singleton, however that adds another layer of complexity when many clients could be queuing messages.

Calling dispose in a deconstructor would certainly help, currently when the client is GC'd the message queue should also be collected alongside it

PR #509, opened 2024-02-05
Convert the GlobalExceptionBridge to use WeakReference for events

### LINKED ISSUE #508: Memory Leak and Stability Concerns Introduced by UnhandledExceptionBridge Subscription in RaygunClientBase Constructor (#504)
**Describe the bug**

Since version 8.1.0 `RaygunClientBase` subscribes to `UnhandledExceptionBridge.OnUnhandledException` in the constructor.
That means that RaygunClients will never be garbage collected, because their reference will be captured in `UnhandledExceptionBridge.OnUnhandledException` event handler.

In the context of AspNetCore, an instance of a raygun client is created for every request when an error occurs ([link to code](https://github.com/MindscapeHQ/raygun4net/blob/6d2678782c13561c118c8b0fe5277c1cf8e5239d/Mindscape.Raygun4Net.AspNetCore/RaygunClientProvider.cs#L21)).

At the very best, this change introduces a memory leak. At the worse it can have a significant impact on platform stability.

Link to the PR and line introducing the change https://github.com/MindscapeHQ/raygun4net/pull/504 ([this change](https://github.com/MindscapeHQ/raygun4net/pull/504/files#diff-e62b402362f36cf71400ff38c33a5280be1127c4eeac632fdb7e15d40cc2c370R102))

**Expected behavior**
What is the expected lifetime of `RaygunClientBase`?
If it's supposed to be a singleton, it shouldn't be created for every request in AspNetCore.
If it's supposed to be transient, it should be disposable to unsubscribe from the event handler.

xenolightning:
This is to fix #508 

Use WeakReference to wrap the event subcribers.

This allows for multiple RaygunClient instances, and they are no longer kept alive by the event subscription.

### DIFF
--- a/CHANGE-LOG.md
+++ b/CHANGE-LOG.md
@@ -1,5 +1,8 @@
 # Full Change Log for Raygun4Net.* packages
 
+### v8.2.1
+- Fixed memory leak when multiple RaygunClient instances are created
+
 ### v8.2.0
 - Added public ctor to allow RaygunClient to accept a custom HttpClient
 - Changed the default timeout of HttpClient from 100 seconds to 30 seconds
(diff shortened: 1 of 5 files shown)


phillip-haydon (dismissed):
:shipit:

phillip-haydon (approved):
:shipit:

Issue #513, opened 2024-02-08
LockRecursionException on UnhandledExceptionBridge on 9.0.0


theolivenbaum:
Just saw this exception when using Raygun4net, seems to be related to [this file](https://github.com/MindscapeHQ/raygun4net/blob/aab8972d224d5a8fd4e15934f0a1e73f1985d765/Mindscape.Raygun4Net.NetCore.Common/UnhandledExceptionBridge.cs#L70).

`````
Unhandled Exception, will terminate now!
System.Threading.LockRecursionException: Write lock may not be acquired with read lock held. This pattern is prone to deadlocks. Please ensure that read locks are released before taking a write lock. If an upgrade is necessary, use an upgrade lock in place of the read lock.
   at System.Threading.ReaderWriterLockSlim.TryEnterWriteLockCore(TimeoutTracker timeout)
   at System.Threading.ReaderWriterLockSlim.EnterWriteLock()
   at Mindscape.Raygun4Net.UnhandledExceptionBridge.RemoveDeadHandler(WeakExceptionHandler handler) in C:\buildAgent\work\75116afedfe31196\Mindscape.Raygun4Net.NetCore.Common\UnhandledExceptionBridge.cs:line 72
   at Mindscape.Raygun4Net.UnhandledExceptionBridge.WeakExceptionHandler.Invoke(Exception exception, Boolean isTerminating) in C:\buildAgent\work\75116afedfe31196\Mindscape.Raygun4Net.NetCore.Common\UnhandledExceptionBridge.cs:line 123
   at Mindscape.Raygun4Net.UnhandledExceptionBridge.RaiseUnhandledException(Exception exception, Boolean isTerminating) in C:\buildAgent\work\75116afedfe31196\Mindscape.Raygun4Net.NetCore.Common\UnhandledExceptionBridge.cs:line 40
   at Mindscape.Raygun4Net.UnhandledExceptionBridge.<>c.<.cctor>b__3_1(Object sender, UnobservedTaskExceptionEventArgs args) in C:\buildAgent\work\75116afedfe31196\Mindscape.Raygun4Net.NetCore.Common\UnhandledExceptionBridge.cs:line 27
   at System.Threading.Tasks.TaskExceptionHolder.Finalize()
`````

theolivenbaum:
Probably introduced by [this change](https://github.com/MindscapeHQ/raygun4net/commit/29303fa7afac1a59ffaef2da202cac7730bc30f0).

[...]

PR #514, opened 2024-02-08
Fix Lock recursion 

xenolightning:
This addresses the Lock Recursion exception raised in #513 

Don't try to clean up the handlers in the invocation.

Instead move all the logic to the add method.

This is slightly less optimal on the memory, but makes managing the locks across threads much simpler and less error prone.

The test was created and was able to reproduce the issue in 9.0.0, the "broken" test now passes.

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Tests/UnhandledExceptionBridgeTests.cs
+++ b/Mindscape.Raygun4Net.NetCore.Tests/UnhandledExceptionBridgeTests.cs
@@ -1,6 +1,4 @@
 using System;
-using System.Threading;
-using System.Threading.Tasks;
 using NUnit.Framework;
 
 namespace Mindscape.Raygun4Net.NetCore.Tests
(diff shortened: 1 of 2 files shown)


Mindscape.Raygun4Net.NetCore.Tests/UnhandledExceptionBridgeTests.cs
```diff
@@ -0,0 +1,47 @@
...
+
+      // GC the callbacks, causing their handlers to be marked as not alive
+      GC.Collect();
+      GC.WaitForPendingFinalizers();
+      GC.Collect();
+      
+      try
+      {
+        // Manually raise an unhandled exception, to handlers that are not alive
+        // Which should cause the lock upgrade exception
```
phillip-haydon, line 36: This comment is really hard to read...

Mindscape.Raygun4Net.NetCore.Tests/UnhandledExceptionBridgeTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Mindscape.Raygun4Net.NetCore.Tests
+{
+  [TestFixture]
+  public class UnhandledExceptionBridgeTests
+  {
+    /// <summary>
+    /// GitHub Issue: 513
+    /// See https://github.com/MindscapeHQ/raygun4net/issues/513 for the issue report
+    /// </summary>
```
phillip-haydon, line 12: 👍

phillip-haydon (approved):
Lets get 9.0.1 out!

Issue #465, opened 2022-03-16
Breadcrumbs not available when upgrading a WinForms app to .NET (Core)


dougwaldron:
I utilize `RaygunClient.RecordBreadcrumb` in a .NET Framework WinForms app that I'm trying to upgrade to .NET 6. 

I've changed the NuGet package from "Mindscape.Raygun4Net" to "Mindscape.Raygun4Net.NetCore". Is this correct?

Now breadcrumbs are no longer available. Are there plans to implement this feature? Or is there an alternative?

FYI, your [documentation for using Breadcrumbs in .NET](https://raygun.com/documentation/language-guides/dotnet/crash-reporting/breadcrumbs/) does not mention that they aren't available in some types of apps.

bounav:
Glad it's not just me!

I mentioned that to Raygun's customer support a few months ago, they confirmed that breadcrumbs are not supported in .net core after a bit of back and forth.

I suggested they make that _clearer_ in their documentation but they haven't got round to doing that it seems.

phillip-haydon:
v10.0.0 is released with breadcrumb support.

PR #516, opened 2024-02-16
Add breadcrumbs to Raygun4Net.NetCore/AspNetCore

ProRedCat:
This brings Breadcrumbs from Raygun4Net4 to the NetCore libraries, some features are removed such as breadcrumb levels, optional stack frame (always included now), and internal error throwing.

- Adds breadcrumbs to Raygun4Net.NetCore/AspNetCore. Uses AsyncLocal with the async context set at the middleware layer for AspNetCore so that breadcrumbs are in scope only for the request they belong to. 
AsyncLocal is the default storage, but it can be set to a global scope (InMemoryBreadcrumbStorage) - Raygun4Maui will use the global scope version.

- Adds packing script to run on MacOS and Linus (translation of existing powershell script) for local development


Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
```diff
@@ -301,6 +301,7 @@ protected override async Task<RaygunMessage> BuildMessage(Exception exception, I
         .SetTags(tags)
         .SetUserCustomData(userCustomData)
         .SetUser(userInfoMessage ?? UserInfo ?? (!String.IsNullOrEmpty(User) ? new RaygunIdentifierMessage(User) : null))
+        .SetBreadcrumbs(Breadcrumbs.Any() ? Breadcrumbs.ToList() : null)
```
xenolightning: Could be simply `Breadcumbs?.ToList()`, it's either empty, got stuff, or null.

Any(), ToList() "enumerates" the collection twice. Even though the Any short circuits it's more efficient and concise doing it just once.

Mindscape.Raygun4Net.NetCore.Common/Breadcrumbs/RaygunBreadcrumb.cs
```diff
@@ -0,0 +1,38 @@
...
+  {
+    public RaygunBreadcrumb()
+    {
+      CustomData = new Dictionary<string, object>();
+      Level = RaygunBreadcrumbLevel.Info;
+      Type = RaygunBreadcrumbType.Manual.ToString();
+      Timestamp = (long)(DateTime.UtcNow - UnixEpoch).TotalMilliseconds;
+    }
+
+    private static readonly DateTime UnixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
xenolightning, line 17: We usually follow this convention for class layout.

const/statics, readonly, fields, properties, ctors, methods

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -32,6 +33,8 @@ public abstract class RaygunClientBase
     private readonly string _apiKey;
     private readonly List<Type> _wrapperExceptions = new List<Type>();
 
+    protected static RaygunBreadcrumbs Breadcrumbs;
```
xenolightning: protected static is a weird combination - not seen too often.

Does it need to be static?

xenolightning (changes requested):
We should also be writing tests to cover usage scenarios

Mindscape.Raygun4Net.NetCore.Common/Breadcrumbs/Enums.cs
```diff
@@ -0,0 +1,15 @@
...
+  {
+    Debug,
+    Info,
+    Warning,
+    Error,
+  }
+
+  public enum RaygunBreadcrumbType
+  {
+    Manual
```
phillip-haydon, line 13: This seems pointless, can just use the word directly in the code for the type for now. As a side you can use `nameof(...)` on the enum instead of `ToString()`, its faster and cleaner.

Mindscape.Raygun4Net.NetCore.Common/Breadcrumbs/Enums.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Mindscape.Raygun4Net.Breadcrumbs
+{
+  public enum RaygunBreadcrumbLevel
```
phillip-haydon, line 3: Can we ditch this entirely for now? Lets not entrance ourselves into `Levels` when we may decide to not support this at all.

xenolightning:
Err I seemed to have commented on an older commit, let me look again

Mindscape.Raygun4Net.AspNetCore/RaygunAspNetMiddleware.cs
```diff
@@ -34,6 +35,11 @@ public async Task Invoke(HttpContext httpContext)
       MemoryStream buffer = null;
       Stream originalRequestBody = null;
 
+      if (RaygunBreadcrumbs.Storage is AsyncLocalBreadcrumbStorage)
```
xenolightning, line 38: I would probably look to change this.

Either create two interfaces `IContextAwareStorage : IStorage` or use a default implementation on the interface. I think the first might be slightly cleaner

Mindscape.Raygun4Net.NetCore.Common/Breadcrumbs/IRaygunBreadcrumbStorage.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+  public interface IRaygunBreadcrumbStorage
+  {
+    void Store(RaygunBreadcrumb breadcrumb);
+
+    void Clear();
+
+    int Size();
+
+    IList<RaygunBreadcrumb> Dump();
```
xenolightning, line 13: Dump was just my human words. The method should probably get `ToList()` or `GetAll()`

Mindscape.Raygun4Net.AspNetCore/RaygunAspNetMiddleware.cs
ProRedCat: Created IContextAwareStorage which provices BeginContext and EndContext. AsyncLocalBreadcrumbStorage implements this now

Mindscape.Raygun4Net.NetCore.Common/Breadcrumbs/Enums.cs
ProRedCat: Levels have now been removed, and the settings in RaygunSettingsBase were removed as well

Mindscape.Raygun4Net.NetCore.Common/Breadcrumbs/Enums.cs
ProRedCat: Removed the enum, it's now hardcoded for manual breadcrumb type

Mindscape.Raygun4Net.NetCore.Common/Breadcrumbs/IRaygunBreadcrumbStorage.cs
ProRedCat: Replaced with ToList

phillip-haydon:
Tagging #465

Mindscape.Raygun4Net.AspNetCore/RaygunAspNetMiddleware.cs
```diff
@@ -34,6 +35,11 @@ public async Task Invoke(HttpContext httpContext)
       MemoryStream buffer = null;
       Stream originalRequestBody = null;
 
+      if (RaygunBreadcrumbs.Storage is IContextAwareStorage storage)
+      {
+        storage.BeginContext();
+      }
+
```
phillip-haydon, line 42: :+1: this is much cleaner than the casting from before.

Mindscape.Raygun4Net.NetCore.Common/Breadcrumbs/RaygunBreadcrumb.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    public string MethodName { get; set; }
+
+    public int? LineNumber { get; set; }
+    
+    public RaygunBreadcrumb()
+    {
+      CustomData = new Dictionary<string, object>();
+      Type = "manual";
+      Timestamp = (long)(DateTime.UtcNow - UnixEpoch).TotalMilliseconds;
```
phillip-haydon, line 32: I'm pretty sure you can assign these directly to the props

```
public IDictionary<string, object> CustomData { get; set; } = new ();
```

Cleans up and removes the need to have the explicit ctor.

Mindscape.Raygun4Net.NetCore.Common/RaygunSettings.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using Mindscape.Raygun4Net.Breadcrumbs;
```
phillip-haydon, line 2: Is this un-used? There's no other changes in this file.

Mindscape.Raygun4Net.NetCore.Tests/RaygunBreadcrumbsTests.cs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    static void InMemoryLowerContext()
+    {
+      RaygunBreadcrumbs.Record("Breadcrumb: in context");
+    }
+      
+  }
+  
+
```
phillip-haydon, line 106: My OCD kicking off, whitespace!

pack-all.sh
phillip-haydon, line 1: 👍 thanks for converting this to a shell script for when we are using macOS/Linux.

Mindscape.Raygun4Net.NetCore.Common/RaygunSettings.cs
ProRedCat: Oops you're right, I had the levels in there beforehand which pulled in that import

phillip-haydon (approved):
Don't forget to update docs to include this new feature

phillip-haydon (approved):
Looking good now, the resizing looks cleaner. Get @xenolightning to cast his eye over it then we can get it in.

xenolightning (approved):
Yeah this looks solid 👍🏼

PR #517, opened 2024-02-19
Remove the usage of LocalThread in .NET Framework versions

phillip-haydon:
Sometimes a null-ref exception occurs when building the message due to the fields being persistent in the RaygunClient when using SendInBackground. 

This PR makes the data scoped to the function and passed through.

Also cleaned up some methods now that the version of C# has been updated, we can use optional parameters.

### DIFF
--- a/CHANGE-LOG.md
+++ b/CHANGE-LOG.md
@@ -1,5 +1,9 @@
 # Full Change Log for Raygun4Net.* packages
 
+### v9.0.2
+- Remove the usage of `ThreadLocal` in the RaygunClient (for .NET Framework)
+  - This does not affect the NetCore version 
+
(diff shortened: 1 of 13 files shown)


xenolightning (changes requested):
The code itself look fine.

I think a covering test would be ideal if possible

phillip-haydon:
Hard to write unit tests because HttpServer doesn't fill in the HttpContext, running it up in IIS Express / IIS to test tho everything seems to be working fine.

xenolightning:
Happy to move forward, due to NETFX being a pain to unit test

xenolightning (dismissed):

xenolightning (approved):

PR #523, opened 2024-03-15
Added support for capturing Environment Variables

phillip-haydon:
New settings to enable capturing environment variables. 

```csharp
builder.Services.AddRaygun(builder.Configuration, settings =>
{
  settings.EnvironmentVariables.Add("USER*");
  settings.EnvironmentVariables.Add("POSH_*");
});
```

Can do:
- Exact Match: `PATH`
- Starts With: `POSH_*`
- Ends With: `*_VALUE`
- Contains: `*_banana_*`

If you enter `*` only it will return nothing as we don't want to capture everything, variables should be opt-in and explicit, not capture everything and potentially leak sensitive data by mistake. 

image

### DIFF
--- a/CHANGE-LOG.md
+++ b/CHANGE-LOG.md
@@ -1,5 +1,11 @@
 # Full Change Log for Raygun4Net.* packages
 
+### v10.1.0
+- Add support for capturing Environment Variables in NetCore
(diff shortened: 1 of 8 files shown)


Deacon-McIntyre (approved):
Very nice, a good quick win!

Deacon-McIntyre (approved):

PR #528, opened 2024-05-05
Add Portable PDB support to Raygun4Net and Raygun4NetCore

xenolightning:
This captures the required debugging information for reversing the stacktrace line numbers and file from the stack trace when a PDB is not included in the build output

Updated payload:

image

### DIFF
--- a/Mindscape.Raygun4Net.Core/Builders/RaygunErrorMessageBuilder.cs
+++ b/Mindscape.Raygun4Net.Core/Builders/RaygunErrorMessageBuilder.cs
@@ -15,7 +15,7 @@ namespace Mindscape.Raygun4Net.Builders
   {
     private static readonly ConcurrentDictionary<string, PEDebugInformation> DebugInformationCache = new();
     public static Func<string, PEReader> AssemblyReaderProvider { get; set; } = PortableExecutableReaderExtensions.GetFileSystemPEReader;
-    
+
     public static RaygunErrorMessage Build(Exception exception)
     {
       RaygunErrorMessage message = new RaygunErrorMessage();
@@ -26,7 +26,7 @@ namespace Mindscape.Raygun4Net.Builders
       message.ClassName = FormatTypeName(exceptionType, true);
 
       message.StackTrace = BuildStackTrace(exception);
-      
+
       if (message.StackTrace != null)
       {
         // If we have a stack trace then grab the images out, and put de-dupe them into an array
@@ -170,12 +170,17 @@ namespace Mindscape.Raygun4Net.Builders
 
     private static IEnumerable<PEDebugInformation> GetDebugInfoForStackFrames(IEnumerable<RaygunErrorStackTraceLineMessage> frames)
     {
+      if (DebugInformationCache.IsEmpty)
+      {
+        return Enumerable.Empty<PEDebugInformation>();
+      }
+
       var imageMap = DebugInformationCache.Values.ToDictionary(k => k.Signature);
       var imageSet = new HashSet<PEDebugInformation>();
-      
+
       foreach (var stackFrame in frames)
       {
-        if (imageMap.TryGetValue(stackFrame.ImageSignature, out var image))
+        if (stackFrame.ImageSignature != null && imageMap.TryGetValue(stackFrame.ImageSignature, out var image))
         {
           imageSet.Add(image);
         }
@@ -183,7 +188,7 @@ namespace Mindscape.Raygun4Net.Builders
 
       return imageSet;
     }
-    
+
     private static PEDebugInformation TryGetDebugInformation(string moduleName)
     {
       if (DebugInformationCache.TryGetValue(moduleName, out var cachedInfo))
--- a/Mindscape.Raygun4Net.Core/Builders/RaygunErrorMessageBuilder.cs
+++ b/Mindscape.Raygun4Net.Core/Builders/RaygunErrorMessageBuilder.cs
@@ -115,8 +115,8 @@ namespace Mindscape.Raygun4Net.Builders
           string file = null;
           string className = null;
           var lineNumber = 0;
-          var ilOffset = -1;
-          var methodToken = -1;
+          var ilOffset = StackFrame.OFFSET_UNKNOWN;
+          var methodToken = StackFrame.OFFSET_UNKNOWN;
 
           try
           {
--- a/Mindscape.Raygun4Net.Core/Builders/RaygunErrorMessageBuilder.cs
+++ b/Mindscape.Raygun4Net.Core/Builders/RaygunErrorMessageBuilder.cs
@@ -124,8 +124,10 @@ namespace Mindscape.Raygun4Net.Builders
             lineNumber = frame.GetFileLineNumber();
             methodName = GenerateMethodName(method);
             className = method.ReflectedType != null ? method.ReflectedType.FullName : "(unknown)";
-
             ilOffset = frame.GetILOffset();
+
+            // This might fail in medium trust environments or for array methods,
+            // so don't crash the entire send process - just move on with what we have
             methodToken = method.MetadataToken;
           }
           catch (Exception ex)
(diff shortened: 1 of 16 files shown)


phillip-haydon (dismissed):
:shipit: 

1Q: the offset is -1, if the try fails and -1 is set but we have other info what happens to the offset? Do we ignore if the offset is -1?

xenolightning:
> :shipit:
> 
> 1Q: the offset is -1, if the try fails and -1 is set but we have other info what happens to the offset? Do we ignore if the offset is -1?

Yeah I wanted a non-sensical value for - what I might do is set it to the constant. Which is equal to -1 :D

https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stackframe.offset_unknown?view=net-8.0

ProRedCat:
I am happy with this, it works for MAUI (iOS and Android tested) with the traces being symbolicated when using Sean's tool

image

One thing we noticed was that if any program is created without PDB files, the IL offsets will cause the symbolicated trace to be off by one line. This is not an issue as it's undefined behaviour to build an app and then build the PDB separately.

xenolightning:
Holding on this just for now, we may need to add more information to the "frame" that we can use to determine what PDB we should be looking at, maybe that's assembly, or some sort of signature - not sure yet

xenolightning:
There is issues using the PEReader in the context of the Android platform. Xamarin has some code to read assemblies out of the blob 

We are going to need to explore this code, and probably port it into Raygun4Net to capture the PDB signatures when targeting android 

https://github.com/xamarin/xamarin-android/tree/main/tools/assembly-store-reader

Mindscape.Raygun4Net.NetCore.Common/Mindscape.Raygun4Net.NetCore.Common.csproj
```diff
@@ -39,4 +39,9 @@
     <None Include="..\LICENSE" Pack="true" PackagePath=""/>
     <None Include="..\128x128-transparent.png" Pack="true" PackagePath="\"/>
   </ItemGroup>
+
+
+  <ItemGroup>
+    <PackageReference Include="System.Reflection.Metadata" Version="6.0.1" />
```
xenolightning, line 45: This will cause our nuget packages to bring in a new reference to this package.

It's fairly ubiquitous - and contains the PE Reader etc...

xenolightning:
Raw report inside raygun, after collecting the new fields. The full pipeline supports these fields. They haven't been reversed in the screenshots, but we have the data to do the reversal!

image
image

Mindscape.Raygun4Net.NetCore.Common/Mindscape.Raygun4Net.NetCore.Common.csproj
xenolightning: It should only be required for netstandard20 and 21

Mindscape.Raygun4Net.NetCore.Common/Builders/RaygunErrorMessageBuilder.cs
```diff
@@ -1,14 +1,20 @@
...
+using System.Reflection.PortableExecutable;
 using System.Text;
+using Mindscape.Raygun4Net.Diagnostics;
 
 namespace Mindscape.Raygun4Net
 {
   public class RaygunErrorMessageBuilder
   {
+    private static readonly ConcurrentDictionary<string, PdbDebugInformation> DebugInformationCache = new();
+    public static Func<string, PEReader> AssemblyReaderProvider { get; set; } = PortableExecutableReaderExtensions.GetFileSystemPEReader;
```
xenolightning, line 16: Intention here is that this is replaced with an android version for MAUI.

This needs to be bundled inside the raygun4maui release though, as our code library doesn't have platform targets

Mindscape.Raygun4Net.Core/Mindscape.Raygun4Net.Core.csproj
```diff
@@ -46,4 +46,8 @@
     <None Include="..\128x128-transparent.png" Pack="true" PackagePath="\" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Reflection.Metadata" Version="6.0.1" />
```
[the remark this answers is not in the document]
xenolightning: NETFX will require it, but I need to fix this for .net core yep

Mindscape.Raygun4Net.NetCore.Common/Mindscape.Raygun4Net.NetCore.Common.csproj
xenolightning: jaaa - I will fix

Mindscape.Raygun4Net.Core/Builders/RaygunErrorMessageBuilder.cs
```diff
@@ -138,5 +163,31 @@ public static RaygunErrorStackTraceLineMessage[] BuildStackTrace(StackTrace stac
...
+          DebugInformationCache.TryAdd(moduleName, debugInfo);
+          return debugInfo;
+        }
+      }
+      catch (Exception ex)
+      {
+        Debug.WriteLine($"Could not load debug information: {ex}");
+      }
+
+      return null;
```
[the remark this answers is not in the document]
xenolightning: Interesting, I originally had this code, as that something went wrong and we shouldn't add it, and maybe we could try again.

But we there's no exceptions, maybe we do add a null?

xenolightning:
Tagged @QuantumNightmare on this one too, as he had suggestions on improving the payload

Mindscape.Raygun4Net.Core/Builders/RaygunErrorMessageBuilder.cs
```diff
@@ -26,6 +26,13 @@ public static RaygunErrorMessage Build(Exception exception)
       message.ClassName = FormatTypeName(exceptionType, true);
 
       message.StackTrace = BuildStackTrace(exception);
+      
+      if (message.StackTrace != null)
+      {
+        // If we have a stack trace then grab the images out, and put de-dupe them into an array
```
QuantumNightmare, line 32: put dedupe?

Mindscape.Raygun4Net.NetCore.Common/Builders/RaygunErrorMessageBuilder.cs
```diff
@@ -199,12 +197,19 @@ public static RaygunErrorMessage Build(Exception exception)
         message.Data = data;
       }
 
+      if (message.StackTrace != null)
+      {
+        // If we have a stack trace then grab the images out, and put de-dupe them into an array
```
QuantumNightmare, line 202: put dedupe

QuantumNightmare (commented):
Last 3 commits look good to me, just commented on what looks like a typo. Would be good to avoid the duplicate code too, but that could be left for another round as I know there are already thoughts on simplifying the repository.

Mindscape.Raygun4Net.Core/Builders/RaygunErrorMessageBuilder.cs
xenolightning: hahaha, brain working faster than fingers

xenolightning:
> Last 3 commits look good to me, just commented on what looks like a typo. Would be good to avoid the duplicate code too, but that could be left for another round as I know there are already thoughts on simplifying the repository.

Yeah, we end up having to make changes to both framework and core sadly.

We do have plans to consolidate them into a single netstandard targeting lib at some point soon

QuantumNightmare (commented):
Thanks for the typo fixes

phillip-haydon (approved):
🚀

phillip-haydon:
Don't forget to update the Change Log.

xenolightning:
> Don't forget to update the Change Log.

I'll do a larger changelog update for 11.0 which will also include the offline storage support for netcore

xenolightning:
Going to merge and release as 11.0-pre1

Then we can base raygun4maui off this package, and bring the rest into line

PR #529, opened 2024-05-09
Fix Issue Where WeakReference Delegate is Garbage Collected While Client is Alive

ProRedCat:
Fixes an issue where the reference to the delegate that sends unhandled exceptions to the client is garbage collected despite the client being alive. This can cause uncaught exceptions not to be reported to Raygun.

This is fixed by adding a strong reference in the client so that while the client is alive the delegate should be alive too.

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Tests/UnhandledExceptionBridgeTests.cs
+++ b/Mindscape.Raygun4Net.NetCore.Tests/UnhandledExceptionBridgeTests.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 
 namespace Mindscape.Raygun4Net.NetCore.Tests
 {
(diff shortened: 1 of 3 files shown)


Mindscape.Raygun4Net.NetCore.Tests/UnhandledExceptionBridgeTests.cs
```diff
@@ -5,6 +5,14 @@ namespace Mindscape.Raygun4Net.NetCore.Tests
   [TestFixture]
   public class UnhandledExceptionBridgeTests
   {
+    private RaygunClientBase _strongRef;
+
+    [SetUp]
+    public void Setup()
+    {
+      _strongRef = new RaygunClient(new RaygunSettings(){CatchUnhandledExceptions = true, ApiKey = "dummy"});
```
xenolightning, line 13: Should fix the formatting here.

Mindscape.Raygun4Net.NetCore.Tests/UnhandledExceptionBridgeTests.cs
```diff
@@ -45,5 +53,25 @@ void Callback(Exception e, bool b)
...
+      };
+
+      // Call GC, this should NOT GC the delegate which the UnhandledExceptionBridge holds for the RaygunClient's 
+      GC.Collect();
+      GC.WaitForPendingFinalizers();
+      GC.Collect();
+      
+      UnhandledExceptionBridge.RaiseUnhandledException(new Exception("Alive"), false);
+      
+      Assert.That(errorMessage, Is.Not.Null);
```
xenolightning, line 74: This test might be prone to a race condition in async code.

We could add a semaphore / signal after the callback has been fired and then check the existence of the error

xenolightning (changes requested):

phillip-haydon (approved):
:+1: was writing to say you need change log but looks like you added it while i was afk :D

xenolightning (approved):
:shipit:

PR #530, opened 2024-05-14
Add support for offline storage of errors to Raygun4Net Core

xenolightning:
Add offline storage of crash reports to Raygun4Net

- This adds a new class `OfflineStoreBase`
- This class can be extended to support any sort of storage of errors, and any custom logic
- We have 2 default store implementations, FileSystem and Local App Data
- New interface `IBackgroundSendStrategy` which is used to determine how often the store should attempt to send it's errors
- By default there is a timer based strategy, which uses a background timer to initiate a check at certain intervals

All these classes are designed to be able to be updated/changed by users if they need specific functionality

MAUI will have it's own store implementation

~This is an early implementation to allow for early feedback and suggestions~

~It is intended that the storage is treated as a singleton. There are default implementations in the net core project, but for other providers such as MAUI there should be another implementation.~

~There is also a singleton background sender, which has an internal timer.~

~Due to the way RaygunClientBase isn't exactly a singleton, we need to loosely couple the actual send method using a delegate (gross)~


xenolightning:
@phillip-haydon this is mostly ready for review.

Can review the approach, and the linkage to the client for now

There is a singleton background worker, and it is wired up to a cache - which should also be a singleton.

The clients (not a singleton yet) can then use the offline store to put failed crash reports, to be picked up by the background worker at a later date.

If it fails to send, then it does not remove the cached error from the disk

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -485,35 +527,70 @@ public async Task Send(RaygunMessage raygunMessage, CancellationToken cancellati
...
+          // Cause an exception to be bubbled up the stack
+          response.EnsureSuccessStatusCode();
         }
       }
       catch (Exception ex)
       {
         Debug.WriteLine($"Error Logging Exception to Raygun {ex.Message}");
 
-        if (_settings.ThrowOnError)
+        if (!hasMessageBeenStored)
```
PanosNB, line 573: Should we move this to finally such that you don't to the offline write call twice?

Mindscape.Raygun4Net.NetCore.Common/Storage/CachedCrashReportBackgroundWorker.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+      Debug.WriteLine($"Exception sending offline errors: {ex}");
+    }
+  }
+
+  /// <summary>
+  /// Start the internal timer. This will enable the sending of any offline stored errors.
+  /// This requires SetSendCallback to be 
+  /// </summary>
+  public static void Start()
```
PanosNB, line 95: Do we need this public? What happens if someone calls it outside the static initializer?

PanosNB: Or if we do need it exposed, should have conditional logic in case _isRunning is already true?

Mindscape.Raygun4Net.NetCore/Storage/FileSystemCrashReportCache.cs
```diff
@@ -0,0 +1,122 @@
...
+  public virtual async Task<List<CrashReportCacheEntry>> GetAll(CancellationToken cancellationToken)
+  {
+    var crashFiles = Directory.GetFiles(_storageDirectory, $"*.{CacheFileExtension}");
+    var errorRecords = new List<CrashReportCacheEntry>();
+
+    foreach (var crashFile in crashFiles)
+    {
+      try
+      {
+        using var fileStream = new FileStream(crashFile, FileMode.Open, FileAccess.Read);
```
PanosNB, line 33: Is there implicit scope for the usings here? Where do they close?

Mindscape.Raygun4Net.NetCore/Storage/FileSystemCrashReportCache.cs
```diff
@@ -0,0 +1,122 @@
...
+  private const string CacheFileExtension = "rgcrash";
+  private readonly string _storageDirectory;
+  private readonly ConcurrentDictionary<Guid, string> _cacheLocationMap = new();
+
+  public FileSystemCrashReportCache(string storageDirectory)
+  {
+    _storageDirectory = storageDirectory;
+  }
+
+  public virtual async Task<List<CrashReportCacheEntry>> GetAll(CancellationToken cancellationToken)
```
PanosNB, line 24: Isn't this too risky of causing an out of memory? Or the very least thrashing the GC? Why do we need them all fetched in memory and we don't just send them to Raygun?

Mindscape.Raygun4Net.NetCore/Storage/FileSystemCrashReportCache.cs
```diff
@@ -0,0 +1,122 @@
...
+      var cacheEntry = new CrashReportCacheEntry
+      {
+        Id = cacheEntryId,
+        ApiKey = apiKey,
+        MessagePayload = payload
+      };
+      var filePath = GetFilePathForCacheEntry(cacheEntryId);
+      var jsonContent = SimpleJson.SerializeObject(cacheEntry);
+
+      using var fileStream = new FileStream(filePath, FileMode.Create);
```
PanosNB, line 71: Is it one file per crash?

PanosNB (commented):
Lot's of good work! Some questions on my end

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -485,35 +527,70 @@ public async Task Send(RaygunMessage raygunMessage, CancellationToken cancellati
...
         {
-          Debug.WriteLine($"Error Logging Exception to Raygun {result.ReasonPhrase}");
+          Debug.WriteLine($"Error Logging Exception to Raygun {response.ReasonPhrase}");
 
-          if (_settings.ThrowOnError)
+          if ((int)response.StatusCode >= 500)
           {
-            throw new Exception("Could not log to Raygun");
+            // If we got a server error then add it to offline storage to send later
+            hasMessageBeenStored = await SaveMessageToOfflineCache(payload, apiKey, cancellationToken);
```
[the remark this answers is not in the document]
xenolightning: The idea here is that we only want to store crashes when raygun is unreachable, which will be any http errors in the 5xx range.

We don't want to store anything from 2xx or 4xx, because that signals raygun was alive, and accepted or rejected the payload.

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
xenolightning: Yeah that might tidy this up a little, I'll take a look.

We need to send when 5xx or when there's a network outage

Mindscape.Raygun4Net.NetCore.Common/Storage/CachedCrashReportBackgroundWorker.cs
xenolightning: I want to give users the ability to turn on/off the background worker if they want to

Mindscape.Raygun4Net.NetCore/Storage/FileSystemCrashReportCache.cs
xenolightning: Yeah that's the idea, it's easier to manage what has/hasn't been sent without loading in entire collections into memory.

We'd want to limit the number of files etc... though

Mindscape.Raygun4Net.NetCore/Storage/FileSystemCrashReportCache.cs
xenolightning: They close at the end of the scope yep

Mindscape.Raygun4Net.NetCore/Storage/FileSystemCrashReportCache.cs
xenolightning: I did think about optimizing this to lazy load / stream the results in, but that could get awkward if there's a lot of files being continuously being created.

Certainly should be altered before making it production ready though

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -485,35 +527,70 @@ public async Task Send(RaygunMessage raygunMessage, CancellationToken cancellati
...
       }
 
+      try
+      {
+        var messagePayload = SimpleJson.SerializeObject(raygunMessage);
+        await SendPayloadAsync(messagePayload, _settings.ApiKey, cancellationToken);
+      }
+      catch (Exception ex)
+      {
+        if (_settings.ThrowOnError)
```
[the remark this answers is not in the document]
xenolightning: It is yeah, it's actually original code, but I changed the method too much the diff doesn't show it well

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -102,31 +106,56 @@ private void OnApplicationUnhandledException(Exception exception, bool isTermina
...
     {
     }
 
+    // ReSharper disable once IntroduceOptionalParameters.Global
     protected RaygunClientBase(RaygunSettingsBase settings, HttpClient client, IRaygunUserProvider userProvider)
+      : this(settings, client, userProvider, null)
+    {
+    }
+
+    protected RaygunClientBase(RaygunSettingsBase settings, HttpClient client, ICrashReportCache crashReportCache)
```
xenolightning, line 136: @phillip-haydon

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -102,31 +106,56 @@ private void OnApplicationUnhandledException(Exception exception, bool isTermina
...
-      
+
       _onUnhandledExceptionDelegate = OnApplicationUnhandledException;
-      
+
       UnhandledExceptionBridge.OnUnhandledException(_onUnhandledExceptionDelegate);
+
+      if (crashReportCache != null)
+      {
+        _crashReportCache = crashReportCache;
```
xenolightning, line 156: @phillip-haydon

Mindscape.Raygun4Net.NetCore.Common/Storage/CachedCrashReportBackgroundWorker.cs
```diff
@@ -0,0 +1,109 @@
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Mindscape.Raygun4Net.Storage;
+
+public static class CachedCrashReportBackgroundWorker
```
xenolightning, line 8: @phillip-haydon

Mindscape.Raygun4Net.NetCore/Storage/FileSystemCrashReportCache.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.IO.Compression;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Mindscape.Raygun4Net.Storage;
+
+public class FileSystemCrashReportCache : ICrashReportCache
```
xenolightning, line 13: @phillip-haydon

xenolightning (commented):
Tagged phill to review the public api surface in specific areas

phillip-haydon (approved):
Only 1 comment..

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
xenolightning: So this is where I started, but the DI extensions is only in the asp.net project, or maui etc...

Not the base library, so it's not always available in it's current form.

Ideally we should have a bunch of services in the container and construct a singleton from the MS DI container, but that's.... a while away

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -485,35 +527,70 @@ public async Task Send(RaygunMessage raygunMessage, CancellationToken cancellati
...
-        var result = await _client.SendAsync(requestMessage, cancellationToken).ConfigureAwait(false);
-
-        if (!result.IsSuccessStatusCode)
+        if (!response.IsSuccessStatusCode)
         {
-          Debug.WriteLine($"Error Logging Exception to Raygun {result.ReasonPhrase}");
+          Debug.WriteLine($"Error Logging Exception to Raygun {response.ReasonPhrase}");
 
-          if (_settings.ThrowOnError)
+          if ((int)response.StatusCode >= 500)
```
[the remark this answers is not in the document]
xenolightning: Status code is an enum, and we can't do gt / lt checks on enum values. 

The enum value is integer backed though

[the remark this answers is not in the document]
xenolightning: I stand corrected! We can do an enum compare hahah.

I've just never done it before, and assumed it needed to be downcasted to an int

Will push a change

Mindscape.Raygun4Net.NetCore.Tests/RaygunMessageBuilderTests.cs
```diff
@@ -239,7 +239,7 @@ public void SetEnvironmentDetails_WithEnvironmentVariables_Star_ShouldReturnNoth
           "*",
           "**",
           "***",
-          "* *",
+          //"* *",
```
xenolightning, line 242: Bleh, need to undo

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -97,6 +97,16 @@ public virtual bool CatchUnhandledExceptions
       set => _settings.CatchUnhandledExceptions = value;
     }
 
+    public ICrashReportCache CrashReportCache
+    {
+      get => _crashReportCache;
+      set
+      {
+        _crashReportCache = value;
+        CachedCrashReportBackgroundWorker.SetSendCallback(SendPayloadAsync);
```
ProRedCat, line 106: Will we have any lifetime issues with this delegate? #529 might have been because of `WeakReference` but just want to make sure that we don't have that happen here

xenolightning: It will keep one client alive at all times, I'm working on changing the way this hooks into the client.

But it won't change the fact we need a single client alive to send the payloads - so sadly yeah, it has to be like this until we look at making the client a singleton, and can have a bit more of a robust link

Mindscape.Raygun4Net.NetCore.Common/Offline/OfflineStoreBase.cs
```diff
@@ -0,0 +1,59 @@
+#nullable  enable
```
PanosNB, line 1: There are two spaces here (Don't be blocked on this, haha)

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -485,35 +508,65 @@ public async Task Send(RaygunMessage raygunMessage, CancellationToken cancellati
...
       }
       catch (Exception ex)
       {
-        Debug.WriteLine($"Error Logging Exception to Raygun {ex.Message}");
+        Debug.WriteLine($"Error Logging Exception to Raygun: {ex.Message}");
 
-        if (_settings.ThrowOnError)
+        // If we got no response or an unexpected server error then add it to offline storage to send later
+        // we get no response if the send call fails for any other reason (network etc)
+        var shouldStoreMessage = response is null || response.StatusCode >= HttpStatusCode.InternalServerError;
```
PanosNB, line 550: Instead of a greater-than-check in here, should we add a static method to assess if the status code is valid? Also, the greater-than check is a bit unintuitive. Should we do not in X?

PanosNB (approved):
All good! A couple of non-blocking comments

Mindscape.Raygun4Net.NetCore.Common/Offline/OfflineStoreBase.cs
xenolightning: Doh, I'll fix it

Mindscape.Raygun4Net.NetCore.Common/Offline/OfflineStoreBase.cs
```diff
@@ -0,0 +1,59 @@
...
+
+  public virtual void SetSendCallback(SendHandler sendHandler)
+  {
+    SendCallback = sendHandler;
+  }
+
+  protected virtual async Task ProcessOfflineCrashReports()
+  {
+    if (SendCallback is null)
+      return;
```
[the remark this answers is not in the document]
xenolightning: Yeah, bad habit, doesn't conform to out styles either, will fix

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
xenolightning: I already check that it's valid on line 542.

That throws an exception if it's invalid

We could check that it's not in 500, 501, 502, 503, 504 ...

The RFC spec says that anything starting with a 5 is a server error, and if that's the case we should store for later, for brevity checking >= is the same as checking the first digit of the response is a 5(xx)

I'll add a comment, but checking a list is more error prone for unexpected 5xx error codes, and provides little value

phillip-haydon (changes requested):
><

Mindscape.Raygun4Net.NetCore/Storage/FileSystemCrashReportStore.cs
```diff
@@ -0,0 +1,132 @@
...
+
+    return errorRecords;
+  }
+
+  public override async Task<bool> Save(string payload, string apiKey, CancellationToken cancellationToken)
+  {
+    var cacheEntryId = Guid.NewGuid();
+    try
+    {
+      Directory.CreateDirectory(_storageDirectory);
```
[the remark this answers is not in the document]
xenolightning: Nah, it just doesn't create a directory if it exists, and returns the directoryinfo of the existing directory

Mindscape.Raygun4Net.NetCore/Storage/LocalApplicationDataCrashReportStore.cs
```diff
@@ -0,0 +1,36 @@
...
+  {
+  }
+
+  private static string GetLocalAppDirectory(string directoryName)
+  {
+    directoryName ??= CreateUniqueDirectory();
+    return Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), directoryName);
+  }
+
+  private static string CreateUniqueDirectory()
```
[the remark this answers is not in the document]
xenolightning: I'll rename the method to directory name :)

Mindscape.Raygun4Net.NetCore/RaygunClient.cs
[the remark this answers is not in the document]
xenolightning: Yes, and the validate api key too

PanosNB (approved):
Looks good to me!

Mindscape.Raygun4Net.NetCore.Common/RaygunSettingsBase.cs
```diff
@@ -62,4 +63,6 @@ public RaygunSettingsBase()
   /// Passing in * will be ignored as we do not want to support collecting 'all' environment variables for security reasons.
   /// </remarks>
   public IList<string> EnvironmentVariables { get; set; } = new List<string>();
+
+  public OfflineStoreBase OfflineStore { get; set; }
```
xenolightning, line 67: @phillip-haydon,

@ProRedCat made a good point in the maui space

This isn't configurable as a "setting", think appsettings.json

Do we want to revisit maybe putting it back into the constructor?

phillip-haydon (approved):

Mindscape.Raygun4Net.NetCore.Common/RaygunSettingsBase.cs
xenolightning: resolution: leave it as a setting for now and just require code config because of the "class" nature of it

PR #531, opened 2024-05-24
Add fully quality module name and remove the android handled

xenolightning:
Fix the module name, to fully qualify the path to load.

That way if the executing directory is different to the module directory, it can load the file correctly.

Also remove the unhandled = true flag from Android

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Common/Platforms/AndroidPlatform.cs
+++ b/Mindscape.Raygun4Net.NetCore.Common/Platforms/AndroidPlatform.cs
@@ -49,7 +49,6 @@ namespace Mindscape.Raygun4Net.Platforms
       {
         var exceptionArgsType = exceptionArgs.GetType();
         exception = (Exception)exceptionArgsType.GetProperty("Exception").GetValue(exceptionArgs);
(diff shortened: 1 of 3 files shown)


Mindscape.Raygun4Net.NetCore.Common/Platforms/AndroidPlatform.cs
```diff
@@ -49,7 +49,6 @@ private static void AndroidUnhandledExceptionHandler(object sender, object excep
       {
         var exceptionArgsType = exceptionArgs.GetType();
         exception = (Exception)exceptionArgsType.GetProperty("Exception").GetValue(exceptionArgs);
-        exceptionArgsType.GetProperty("Handled")?.SetValue(exceptionArgs, true);
```
xenolightning, line 52: We don't actually need this, this just prevents the app from crashing on android.

If this is desired behaviour the user can hook this and do it in their code, but we shouldn't be affecting functionality

ProRedCat (approved):
:shipit:

xenolightning:
Released as 11.0.0-pre2

PR #534, opened 2024-06-13
Add `ConfigureAwait(false)` to the blocking and non blocking Send calls

xenolightning:
`GetAwaiter().GetResult()` is really not ideal, we have to continue to support the synchronous api, so it's required.

Internally we should be avoiding the usage of Send()

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
+++ b/Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
@@ -289,6 +289,7 @@ namespace Mindscape.Raygun4Net
     /// Transmits an exception to Raygun synchronously.
     /// </summary>
     /// <param name="exception">The exception to deliver.</param>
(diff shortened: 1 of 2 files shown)


Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -316,7 +316,7 @@ public void Send(Exception exception, IList<string> tags, IDictionary userCustom
     {
       try
       {
-        SendAsync(exception, tags, userCustomData).GetAwaiter().GetResult();
+        SendAsync(exception, tags, userCustomData).ConfigureAwait(false).GetAwaiter().GetResult();
```
xenolightning, line 319: This is the root cause of the deadlock when calling `Send()` on a UI Thread

phillip-haydon (approved):
image

phillip-haydon (approved):
image

phillip-haydon (approved):
image

phillip-haydon:
It's your change log :D

PR #537, opened 2024-07-19
Use Singleton RaygunClient in RaygunHttpModule (.NET Framework)

MattByers:
# Description
Update the `RaygunHttpModule` for ASP.NET Framework applications to store and use a single `RaygunClient` instance. This avoids a new `RaygunClient` and in turn a `ThrottledBackgroundMessageProcessor`, from being created for every exception handled by the `RaygunHttpModule`. The `ThrottledBackgroundMessageProcessor` creates a number of background worker threads, so instantiating it for each exception can be costly.

### DIFF
--- a/Mindscape.Raygun4Net4/RaygunHttpModule.cs
+++ b/Mindscape.Raygun4Net4/RaygunHttpModule.cs
@@ -75,12 +75,7 @@ namespace Mindscape.Raygun4Net
         return raygunApplication.GenerateRaygunClient();
       }
 
-      return _raygunClient ??= GenerateDefaultRaygunClient();
-    }
-
-    private static RaygunClient GenerateDefaultRaygunClient()
-    {
-      return new RaygunClient();
+      return _raygunClient ??= new RaygunClient();
     }
 
     protected bool CanSend(Exception exception)
--- a/Mindscape.Raygun4Net4/RaygunHttpModule.cs
+++ b/Mindscape.Raygun4Net4/RaygunHttpModule.cs
@@ -8,6 +8,8 @@ namespace Mindscape.Raygun4Net
 {
   public class RaygunHttpModule : IHttpModule
   {
+    private RaygunClient _raygunClient;
+
     private bool ExcludeErrorsBasedOnHttpStatusCode { get; set; }
 
     private bool ExcludeErrorsFromLocal { get; set; }
@@ -68,11 +70,17 @@ namespace Mindscape.Raygun4Net
 
     protected RaygunClient GetRaygunClient(HttpApplication application)
     {
-      var raygunApplication = application as IRaygunApplication;
-      return raygunApplication != null ? raygunApplication.GenerateRaygunClient() : GenerateDefaultRaygunClient(application);
+      if (_raygunClient == null)
+      {
+        _raygunClient = application is IRaygunApplication raygunApplication
+          ? raygunApplication.GenerateRaygunClient()
+          : GenerateDefaultRaygunClient();
+      }
+
+      return _raygunClient;
     }
 
-    private RaygunClient GenerateDefaultRaygunClient(HttpApplication application)
+    private static RaygunClient GenerateDefaultRaygunClient()
     {
       var instance = new RaygunClient();
 
--- a/Mindscape.Raygun4Net4/RaygunHttpModule.cs
+++ b/Mindscape.Raygun4Net4/RaygunHttpModule.cs
@@ -82,14 +82,7 @@ namespace Mindscape.Raygun4Net
 
     private static RaygunClient GenerateDefaultRaygunClient()
     {
-      var instance = new RaygunClient();
-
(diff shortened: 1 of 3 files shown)


Mindscape.Raygun4Net4/RaygunHttpModule.cs
```diff
@@ -68,20 +70,19 @@ public void SendError(HttpApplication application, Exception exception)
...
-    {
-      var instance = new RaygunClient();
-
-      if (HttpContext.Current != null && HttpContext.Current.Session != null)
+      if (_raygunClient == null)
       {
-        instance.ContextId = HttpContext.Current.Session.SessionID;
+        _raygunClient = application is IRaygunApplication raygunApplication
+          ? raygunApplication.GenerateRaygunClient()
+          : GenerateDefaultRaygunClient();
```
QuantumNightmare, line 77: The behaviour of providing your own Raygun client has changed here. I wonder if there are some cases where customers are relying on raygunApplication.GenerateRaygunClient being called on every exception. Perhaps to change some setting based on the exception or what user is currently logged in.
As such, perhaps raygunApplication.GenerateRaygunClient should be called every time when present, and we only cache our default RaygunClient as a singleton.

Mindscape.Raygun4Net4/RaygunHttpModule.cs
```diff
@@ -68,20 +70,19 @@ public void SendError(HttpApplication application, Exception exception)
...
-    }
-
-    private RaygunClient GenerateDefaultRaygunClient(HttpApplication application)
-    {
-      var instance = new RaygunClient();
-
-      if (HttpContext.Current != null && HttpContext.Current.Session != null)
+      if (_raygunClient == null)
       {
-        instance.ContextId = HttpContext.Current.Session.SessionID;
```
QuantumNightmare, line 81: Need to be careful when changing where this gets called. I don't know for sure, but we may have been calling it here such that we can cache the contextId before performing any async operations - after which maybe we can't get it anymore. As such, make sure to test that getting this contextId still works in situations where the background/async methods are being used to send the crash report.

QuantumNightmare (commented):
Looking good, just a couple of thoughts to discuss

Mindscape.Raygun4Net4/RaygunHttpModule.cs
phillip-haydon: There have been a couple of questions around this and my reply was to implement `SendingMessage` (i forget the exact name in the netcore version), but would like an example if this doesn't work and so far not had any examples.

QuantumNightmare: How do you attach the OnSendingMessage event hook on the RaygunClient that's created within the HttpModule though? I believe that's why this GenerateRaygunClient method exists.

Mindscape.Raygun4Net4/RaygunHttpModule.cs
QuantumNightmare: I do see that you're pulling the contextId to the side in the StripAndSendInBackground before then setting it again inside the lambda - so maybe that caters for this thought.

Mindscape.Raygun4Net4/RaygunHttpModule.cs
phillip-haydon: Spoke in person to discuss this and conclude I agree with the original statement that `raygunApplication.GenerateRaygunClient` should be called every time. They can cache their own version in there if they need to and allows them to hopefully preserve any functionality they depend on.

MattByers: That's a great point! I'm updated this now.

Mindscape.Raygun4Net4/RaygunHttpModule.cs
QuantumNightmare: Excellent :+1:

Mindscape.Raygun4Net4/RaygunHttpModule.cs
```diff
@@ -68,20 +70,17 @@ public void SendError(HttpApplication application, Exception exception)
...
+        return raygunApplication.GenerateRaygunClient();
       }
 
-      return instance;
+      return _raygunClient ??= GenerateDefaultRaygunClient();
+    }
+
+    private static RaygunClient GenerateDefaultRaygunClient()
+    {
+      return new RaygunClient();
```
QuantumNightmare, line 83: Probably no point having this in a separate method.

QuantumNightmare (approved):
Thanks for the change. Looks good, just comment on one optional minor thing that I didn't think of before.

QuantumNightmare (approved):
Thanks :+1:

QuantumNightmare (approved):
:+1:

Workshop2:
Hey @MattByers - when will this be available in a release?

PR #539, opened 2024-08-06
fix: PEDebugInformation properties as nullable

miquelbeltran:
Part of the solution for the issue reported here: https://raygun.com/forums/thread/186975#187021

As discussed internally on Slack, when running a Raygun4MAUI client, we saw an issue in which the null check for the `Signature` property was not correctly done, and we had the suspicion that the compiler was optimizing the code because `Signature` was not declared as nullable, so the null check may be removed.

So I tried to set this as nullable, packaged a set of `nupkg` using the `pack-all` script, and imported them locally in the Raygun4MAUI solution. Then compiled and run the project (both directly in "debug" and as a "release" APK), and I couldn't reproduce the issue anymore following the same steps.

cc. @ProRedCat @PanosNB


miquelbeltran:
Seems that this won't be necessary anymore, as discussed on Slack.

Issue #536, opened 2024-06-25
RaygunClient.SendInBackground (AspNetCore) and IRaygunUserProvider missing HttpContext


isak1080:
**Describe the bug**
While upgrading to Raygun4net 11 from 9.x I had to move away from `IRaygunClientProvider` to something else.

I now setup my own `IRaygunUserProvider` which uses `IHttpContextAccessor` to grap the user info from the HTTP request. 

My RaygunClient correctly gets my custom user provider in its constructor. 

The problem is when I call `RaygunClient.SendInBackground` - the call to `IRaygunUserProvider.GetUser` is done on a worker thread, which means it cannot get the HttpContext.

How is this supposed to work? My current workaround is to simply build the message myself using the `RaygunMessageBuilder` but then what's the point of doing all this setup with user provider etc.?

phillip-haydon:
Hello!

The message is built before being queued and sent so it should always have access to the HttpContext. 

Are you able to show some code or explain how I ca. reproduce the issue you’re running into?

Thank you.

isak1080:
That's not what I'm seeing. GetUser is called on a worker thread thanks to `Task.Run`

![image](https://github.com/MindscapeHQ/raygun4net/assets/105219900/f43d7d8c-625d-4d6d-a47a-30a244120b21)

See the attached visual studio project that reproduces the issue:
[RaygunClientHttpContextBug.zip](https://github.com/user-attachments/files/15970436/RaygunClientHttpContextBug.zip)

phillip-haydon:
Thank you for the project reproducing the issue, its helpful!

So we will prob need to tweak things a bit to resolve this, similar to what we do for getting the request/response information up front before passing it on.

There's a couple of work arounds I can offer in the short term:

1) Use `.SendAsync(...)`

This builds the message and sends it right away so it doesn't lose context.

2) Use the middleware, when you don't want to capture explicitly you can use the middleware which will pass the http context to SendInBackground. 

3) Use the SendInBackground the middleware uses which causes the message to be built up front.

```
app.MapGet("/crash", (RaygunClient client, IHttpContextAccessor accessor) =>
{
    try
    {
        return 100 / int.Parse("0");
    }
    catch (Exception ex)
    {
        client.SendInBackground(ex, new List<string>(), accessor.HttpContext);
        return -1;
    }
});
```

-----

[...]

PR #540, opened 2024-08-13
Build message before queuing it

phillip-haydon:
This change is required for updates to Serilog provider where Serilog uses the base implementation of `SendInBackground` but depends on `IRaygunUserProvider` to get the user info from `HttpContextAccessor.HttpContext`

When the build message is sent as a delegate to `Enqueue` it gets executed out of context and the information is lost. 

There is a minor overhead in memory building the message up front but this only occurs when there's a huge number of exceptions being thrown.

### DIFF
--- a/CHANGE-LOG.md
+++ b/CHANGE-LOG.md
@@ -1,5 +1,9 @@
 # Full Change Log for Raygun4Net.* packages
 
+### v11.0.4
+- Fix issue with `RaygunClientBase` where `SendInBackground` deferred building the message until late, losing HttpContext
+  - See: https://github.com/MindscapeHQ/raygun4net/pull/540
+
 ### v11.0.3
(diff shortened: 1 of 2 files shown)


sumitramanga (approved):
Code LGTM 🌴

sumitramanga:
How much are we talking in terms of " huge number of exceptions" for the overhead issue?

phillip-haydon:
> How much are we talking in terms of " huge number of exceptions" for the overhead issue?

When we tested it, we chucked it in a loop sending 100k exceptions to see what happened with the memory, it grew a little bit more than if it was delegated, not enough to cause a problem, but just didn't want people to think we were having a memory leak.

This change is more important than ~10mb of data showing up in memory profiling though.

phillip-haydon:
Possibly fixes #536

PR #545, opened 2024-09-16
Removed legacy Xamarin projects

phillip-haydon:
These projects are very old Xamarin projects that are not longer supported.

Backed up into tag: https://github.com/MindscapeHQ/raygun4net/tree/legacy-xamarin-snapshot


ProRedCat (approved):
Hugeeeee

ProRedCat:
With the tag we should update the link on the Xamarin pages on our blogs to direct to that

PR #548, opened 2024-11-05
Change UnobservedTaskException from being tagged as UnhandledException to UnobservedTaskException

ProRedCat:


Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -94,14 +94,14 @@ public virtual bool CatchUnhandledExceptions
...
-    private void OnApplicationUnhandledException(Exception exception, bool isTerminating)
+    private void OnApplicationUnhandledException(Exception exception, bool isTerminating, string tag)
     {
       if (!_settings.CatchUnhandledExceptions)
       {
         return;
       }
 
-      Send(exception, UnhandledExceptionTags);
+      Send(exception, new List<string> { tag });
```
QuantumNightmare, line 104: Looks like the constant UnhandledExceptionTags is no longer used because of this change and so should be deleted too.

Mindscape.Raygun4Net.NetCore.Common/UnhandledExceptionBridge.cs
```diff
@@ -82,15 +82,15 @@ public WeakExceptionHandler(UnhandledExceptionHandler handler)
         _reference = new WeakReference<UnhandledExceptionHandler>(handler);
       }
 
-      public void Invoke(Exception exception, bool isTerminating)
+      public void Invoke(Exception exception, bool isTerminating, string tag)
```
QuantumNightmare, line 85: Perhaps this class is not intended to be used by customers, but this is technically a breaking change due to it being public. (Same as all the public methods above). Shall we instead make an overload for this method and change the other methods above to not have a tag passed in (if that argument is never used) and just pass the string literal from with their method bodies?

QuantumNightmare (changes requested):
A couple of comments to consider

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
ProRedCat: Had that in mind, kept it in case we needed it but likely we do not anymore

Mindscape.Raygun4Net.NetCore.Common/UnhandledExceptionBridge.cs
ProRedCat: Should be addressed now, let me know what you think

Mindscape.Raygun4Net.NetCore.Common/UnhandledExceptionBridge.cs
```diff
@@ -23,22 +24,22 @@ static UnhandledExceptionBridge()
         RaiseUnhandledException(args.ExceptionObject as Exception, args.IsTerminating);
       };
 
-      TaskScheduler.UnobservedTaskException += (sender, args) => { RaiseUnhandledException(args.Exception, false); };
+      TaskScheduler.UnobservedTaskException += (sender, args) => { RaiseUnhandledException(args.Exception,false, "UnobservedTaskException"); };
```
velocitysystems, line 27: @ProRedCat I believe `UnobservedTaskException` may always (or _nearly_ always) raise an `AggregateException`. 
If there is only one inner exception this should _really_ be unboxed. This appears to be the current behavior of the MAUI Xamarin provider.

velocitysystems: This change looks good @ProRedCat. What about `AggregateException`'s which are _not_ unobserved task exceptions (handled + unhandled) - Will they also be unwrapped?

ProRedCat: In this case right now it's only the UnobservedTaskExceptions, will have a chat with some other around handled + unhandled

Mindscape.Raygun4Net.NetCore.Common/UnhandledExceptionBridge.cs
QuantumNightmare: Ah yes, I didn't see this code is in a nested private class

QuantumNightmare (approved):
Thanks for the changes

Mindscape.Raygun4Net.NetCore.Common/UnhandledExceptionBridge.cs
ProRedCat: Observability for others: We do have a feature for automatically unwrapping exceptions.

This is done through `_raygunClient.AddWrapperExtensions(typeof(AggregateException))`

Issue #547, opened 2024-10-12
Access violation - WinRT exception serialization (WindowsAppSdk)


xpetrik:
**Describe the bug**
Crash occurs during the serialization of a WinRT exception using SimpleJson

System.AccessViolationException: 'Attempted to read or write protected memory. This is often an indication that other memory is corrupt.'

**To Reproduce**
1. Demo project based on https://github.com/microsoft/WindowsAppSDK-Samples/tree/main/Samples/AppLifecycle/Restart
2. Project contains sources from your project with modification of visbility of method BuildMessage (public)
3. Folowing code crash on AccessViolatiaon (this is only the essential part of the method to send the report)
```csharp
var raygunClient = new RaygunClient("apikey");              
var message = await raygunClient.BuildMessage(ex, null);
string payload = SimpleJson.SerializeObject(message);
```

**Expected behavior**
no crash and send report

**Screenshots**
![image](https://github.com/user-attachments/assets/c2e0741b-a6f6-4bc8-a666-d985f9c033ca)

**Desktop:**
 - OS: Windows 11 23H2 x64
 - Visual Studio 2022

[cs-winui-packaged.zip](https://github.com/user-attachments/files/17349736/cs-winui-packaged.zip)

QuantumNightmare:
Thanks for reporting this. Unfortunately I have not been able to get your repo project running so far.

I have not worked with WinUI before, so I first installed the necessary components for Visual Studio 2022. When attempting to run, it complained about the WindowsSdkPackageVersion and that the auto generated files contained unsafe code. I also had to delete the win-arm64 profile and the launchsettings file. After all this, it compiles and runs, but then encounters `System.Runtime.InteropServices.COMException: 'Class not registered (0x80040154 (REGDB_E_CLASSNOTREG))'`

So I then created a new WinUI packaged project which compiled and ran just fine. I copied over the Raygun4Net code from your project and the 3 lines of usage code that you provided, which resulted in correctly serializing the message.

Do you know the key differences between the repo you provided and a brand new WinUI packaged project? I have provided my project below in case it helps. The project is called WinUIErrorHarness. The solution contains your original repo along with the modifications I made in an attempt to get it running on my machine.

Is the screenshot you provided within the bowels of the serialization process? Do you know what object and property is trying to be accessed at that point?

[WinUIErrorHarness.zip](https://github.com/user-attachments/files/17357935/WinUIErrorHarness.zip)

xpetrik:
[...]

PR #549, opened 2024-11-12
Filter out __RestrictedErrorObjectReference

QuantumNightmare:
Addresses https://github.com/MindscapeHQ/raygun4net/issues/547

WinRT exceptions can have a value in the Exception.Data dictionary with a key of "__RestrictedErrorObjectReference" and a value of type WinRT.ObjectReferenceWithContext<WinRT.Interop.IUnknownVftbl>. When trying to serialize this value, a System.AccessViolationException is thrown which can crash the application, even if caught.

This PR is a candidate fix for this issue which take the approach of recognizing the Data dictionary key and not including the value in the report sent to Raygun.


QuantumNightmare:
@phillip-haydon Thought I'd open this for discussion first. Any thoughts on this approach to solving the linked issue? I'll bump the version later once we're agreed on a solution.

phillip-haydon (approved):
:+1:

PR #550, opened 2024-11-24
json

Dman0808:


sumitramanga:
Hi @Dman0808 ! 

This is not a valid PR to us, nor does it contain a description of why you are making these changes. One file also has `[` which appears to be a mistake. Please ensure you provide a description on why you are making changes and what value it brings into the provider.

Cheers,
Sumtira

PR #552, opened 2024-11-29
Fix field creation that doesn't currently build using new syntax

phillip-haydon:


ProRedCat (approved):
Language version issue?

PR #553, opened 2024-12-08
Add MS Logger support

### LINKED ISSUE #526: Add default tags to outgoing errors
**Is your feature request related to a problem? Please describe.**
With the deprecation of the following code, we can no longer hook centralized handlers on the `SendingMessage` event easily.
We used this approach to enrich all outgoing messages with tags that captured environment info, like service name, instance, etc.

```csharp
builder.Services.AddRaygun(builder.Configuration, new RaygunMiddlewareSettings
{
  // adds an optional example of over riding the client provider
  ClientProvider = new ExampleRaygunAspNetCoreClientProvider()
});
```

**Describe the solution you'd like**
Maybe an approach similar to `IRaygunUserProvider` or just extending RaygunSettings to allow for default tag list or configure event handlers directly on DI setup.

**Describe alternatives you've considered**
We considered getting the `RaygunClient` singleton from the DI container and adding the `SendingMessage` event handler, but it seems to be a bit of a hack.

phillip-haydon:
New Mindscape.Raygun4Net.Extensions.Logger for supporting MSLogger.

Introduces a new IMessageBuilder to allow customisation of messages without needing api changes or code changes to support things. 

This should fix #526 which would allow adding and registering a new class like:

```csharp
public class DefaultTags : IMessageBuilder
{
  public Task<RaygunMessage> Apply(RaygunMessage message, Exception exception)
  {
    message.Details.Tags.Add("DefaultTag");
    
    return Task.FromResult(message);
  }
}
```

And registering it during startup

```csharp
builder.Services.AddRaygun(settings =>
{
  settings.ApiKey = "*your_api_key*";
});
builder.Services.AddSingleton<IMessageBuilder, DefaultTags>();
```

Which would add the tag during the message building and submit with the default tag.

See Mindscape.Raygun4Net.Extensions.Logging/README.md for MSLogger info.


.editorconfig
```diff
@@ -1,3 +1,14 @@
 [*]
 indent_style = space
-indent_size = 2
\ No newline at end of file
+indent_size = 2
+
+[*.cs]
+# https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0290
+csharp_style_prefer_primary_constructors = false
+dotnet_diagnostic.IDE0290.severity = none
```
phillip-haydon, line 8: This disables suggesting primary constructors.

.editorconfig
```diff
@@ -1,3 +1,14 @@
...
+
+# https://www.jetbrains.com/help/rider/ConvertToPrimaryConstructor.html
+resharper_convert_to_primary_constructor_highlighting = none
+
+# Ensure if statements always use braces
+csharp_prefer_braces = true:error
```
phillip-haydon, line 14: Enforces `if` statements to have braces.

Mindscape.Raygun4Net.AspNetCore/Builders/RaygunAspNetCoreResponseMessageBuilder.cs
```diff
@@ -8,19 +9,19 @@ namespace Mindscape.Raygun4Net.AspNetCore.Builders
   // ReSharper disable once ClassNeverInstantiated.Global
   public class RaygunAspNetCoreResponseMessageBuilder
   {
-    public static RaygunResponseMessage Build(HttpContext? context)
+    public static Task<RaygunResponseMessage> Build(HttpContext? context, RaygunSettings _)
```
phillip-haydon, line 12: No one should be accessing this except us, this is made async for consistency with the Request but will likely be moved into the message builder in the future.

Mindscape.Raygun4Net.Extensions.Logging/RaygunLogger.cs
```diff
@@ -0,0 +1,174 @@
...
+  
+  private async Task SendLogAsync(Exception? exception, List<string> tags, Dictionary<string, string> customData)
+  {
+    try
+    {
+      await _client.SendInBackground(exception, tags, customData);
+    }
+    catch (Exception ex)
+    {
+      // Log the failure to send to Raygun - you might want to use a fallback logger here
```
ProRedCat, line 96: Are these comments for us?

Mindscape.Raygun4Net.AspNetCore.Tests/Mindscape.Raygun4Net.AspNetCore.Tests.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
PanosNB, line 3: This is backwards compatible to 6?

Mindscape.Raygun4Net.AspNetCore/RaygunClient.cs
```diff
@@ -29,11 +29,11 @@ public RaygunClient(RaygunSettings settings, HttpClient httpClient) : base(setti
   {
   }
 
-  public RaygunClient(RaygunSettings settings, IRaygunUserProvider userProvider) : base(settings, userProvider)
+  public RaygunClient(RaygunSettings settings, IRaygunUserProvider userProvider, IEnumerable<IMessageBuilder> messageBuilders) : base(settings, userProvider, messageBuilders)
```
PanosNB, line 32: So we no longer allow a RaygunClient to be constructed with `RaygunClient(RaygunSettings settings, IRaygunUserProvider userProvider)`? Doesn't this introduce a breaking change?

Mindscape.Raygun4Net.Extensions.Logging/RaygunLogger.cs
```diff
@@ -0,0 +1,174 @@
...
+  /// If the log level is below the LogLevel setting, or the OnlyLogExceptions is true, the log will be ignored.
+  /// </remarks>
+  public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
+  {
+    if (!IsEnabled(logLevel) || (exception == null && _settings.OnlyLogExceptions))
+    {
+      return;
+    }
+    
+    var message = formatter?.Invoke(state, exception);
```
PanosNB, line 51: Can we enforce non-null of the formatter parameter type instead of doing the null check here?

Mindscape.Raygun4Net.Extensions.Logging/RaygunLoggerExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+  public static ILoggingBuilder AddRaygunLogger(this ILoggingBuilder builder, IConfiguration? configuration, Action<RaygunLoggerSettings>? options = null)
+  {
+    // Since we are not using IConfiguration, we need to create a new instance of RaygunSettings
+    var settings = new RaygunLoggerSettings();
+
+    // Fetch settings from configuration or use default settings
+    configuration?.GetSection("RaygunSettings").Bind(settings);
+
+    // Override settings with user-provided settings
+    options?.Invoke(settings);
```
PanosNB, line 41: What happens is configuration or options is passed as `null`?

Mindscape.Raygun4Net.Extensions.Logging/RaygunLoggerProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+
+  /// <summary>
+  /// Creates or retrieves a RaygunLogger instance for the specified category.
+  /// </summary>
+  /// <param name="categoryName">The category name for the logger.</param>
+  /// <returns>An ILogger instance configured for the specified category.</returns>
+  /// <exception cref="ObjectDisposedException">Thrown when the provider has been disposed.</exception>
+  public ILogger CreateLogger(string categoryName)
+  {
+    if (_disposed)
```
PanosNB, line 38: Why don't we care about concurrent access to `_disposed` by someone calling `CreateLogger` and `Dispose` (below) about the same time?

Mindscape.Raygun4Net.Extensions.Logging/RaygunLogger.cs
phillip-haydon: Just github co-pilot putting weird shit in

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
```diff
@@ -103,23 +105,25 @@ private void OnApplicationUnhandledException(Exception exception, bool isTermina
     }
 
     protected RaygunClientBase(RaygunSettingsBase settings)
-      : this(settings, DefaultClient, null)
+      : this(settings, DefaultClient, null, [])
```
PanosNB, line 108: Double check my concerns above about breaking changes

Mindscape.Raygun4Net.NetCore/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+    return services;
+  }
+  
+  /// <summary>
+  /// Registers a custom User Provider with the DI container. This allows you to provide your own implementation of IRaygunUserProvider.
+  /// </summary>
+  public static IServiceCollection AddRaygunUserProvider<T>(this IServiceCollection services) where T : class, IRaygunUserProvider
+  {
+    // In case the default or any other user provider is already registered, remove it first
+    var existing = services.FirstOrDefault(x => x.ServiceType == typeof(IRaygunUserProvider));
```
PanosNB, line 54: Does this need to clear more than than one?

PanosNB (changes requested):
Great work! I have some concerns about introducing a breaking change by expecting the RaygunClient constructor to have that extra field. We shouldn't be introducing any, so let me know if I am mistaken :-)

Mindscape.Raygun4Net.NetCore/ApplicationBuilderExtensions.cs
phillip-haydon: I'm unsure, what is not clear about it?

Mindscape.Raygun4Net.NetCore.Common/RaygunClientBase.cs
phillip-haydon: This is calling the base-ctor, and its just adding a default so the main ctors pass it down

Mindscape.Raygun4Net.Extensions.Logging/RaygunLogger.cs
phillip-haydon: Cannot, its not our interface, this comes from MS.

Mindscape.Raygun4Net.AspNetCore.Tests/Mindscape.Raygun4Net.AspNetCore.Tests.csproj
phillip-haydon: Test project, so it doesn't matter.

Mindscape.Raygun4Net.Extensions.Logging/RaygunLoggerExtensions.cs
phillip-haydon: Then the base settings will be used. The new instance of the settings variable defines the default values, these are overridden by the configuration settings from appsettings.json, and further overridden by code level configuration bassed in as ōptions.

Mindscape.Raygun4Net.NetCore/ApplicationBuilderExtensions.cs
phillip-haydon: We only want to ensure 1 is registered, so we can safely assume the one registered by the customer is the one used.

If they register outside of this helper, multiple implementations can be registered. When resolved it will take the last registration, so depending on if they register before or after us would dictate whose gets used.

PanosNB (approved):
This maintains backwards compatibility with the constructors, so all good on my end!

PanosNB (approved):
lgtm

PR #557, opened 2025-01-07
[Cleanup] Narrow target frameworks for tests and ensure we have braces

phillip-haydon:
Make sure the builds run in TeamCity/Linux, and enforces braces.


ProRedCat (approved):
Thanks for fixing the builds here

PR #559, opened 2025-02-13
fix: Add DynamicDependency declaration to preserve MarshalManagedException in iOS and MacOS apps

miquelbeltran:
As discussed on Slack, a customer has been facing issues where the method `MarshalManagedException` gets trimmed by the compiler, so crashes aren't captured.

The first approach I tried was to declare linker rules via xml configuration, but that only works if declared in the root project.

In order to offer a turn-key solution, I've found out that you can also declare the `DynamicDependency` directly in the class code. This will tell the compiler to not trim it. I have added it for both iOS and MacCatalyst platforms.

This has been tested using the provided sample code by the customer.

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Common/Platforms/ApplePlatform.cs
+++ b/Mindscape.Raygun4Net.NetCore.Common/Platforms/ApplePlatform.cs
@@ -13,10 +13,15 @@ namespace Mindscape.Raygun4Net.Platforms
 
     private static object MarshalManagedExceptionMode_UnwindNativeCode;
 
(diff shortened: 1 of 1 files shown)


sumitramanga (approved):
LGTM, can't spot anything wrong with this work :) Nice work!

ProRedCat:
We should add `[DynamicDependency]` on Android as well since we also use reflection to attach the listeners as well there

miquelbeltran:
> We should add `[DynamicDependency]` on Android as well since we also use reflection to attach the listeners as well there

Yes, I was thinking the same. I will do that in a new PR

PR #560, opened 2025-02-17
fix: add DynamicDependency definition for Windows and Android

miquelbeltran:
This PR adds the dynamic dependency definition for Android and Windows platform implementations. However, I am not sure how to effectively test this, I asked on Slack about which existing example project I can use for that, otherwise I will create new example projects.


miquelbeltran:
Tested this on Android (with a sample project from https://github.com/dotnet/android-samples/tree/main) still have to verify it with a Windows app

miquelbeltran:
Tested now on a WinUI Windows app I created here: https://github.com/MindscapeHQ/raygun4net-examples/pull/2

sumitramanga (approved):
Looks good!

PR #565, opened 2025-04-15
docs: Create RELEASING.md

miquelbeltran:
Documents the release process for Raygun4Net

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,77 @@
+# Releasing Raygun4Net
+
+This release process is currently under review: https://github.com/MindscapeHQ/raygun4net/issues/561
+
+Raygun for Net is published on nuget.org under the [mindscapehq account](https://www.nuget.org/profiles/mindscapehq).
+
+## Semantic versioning
+
+This package follows semantic versioning.
+
+Given a version number MAJOR.MINOR.PATCH (x.y.z), increment the:
+
+- MAJOR version when you make incompatible changes
+- MINOR version when you add functionality in a backward compatible manner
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -0,0 +1,77 @@
...
+- `version`: Version is the current version for the release that is signed into the assembly, regardless of the Package
+Version. i.e. `8.1.0`
+
+- `packageVersion`: `PackageVersion` is used by NuGet, this should be the same as `version` except you can change it if
+you need to, i.e. to include `-pre-1` to indicate that `8.1.0` is prerelease version 1. Enter in `8.1.0-pre-1` when
+prompted.
+
+## Preparing for release
+
+To build releases, you need access to Raygun's internal TeamCity instance.
```
sumitramanga, line 37: Not sure if we want to expose internal systems. let's check with the team on this

RELEASING.md
```diff
@@ -0,0 +1,77 @@
+# Releasing Raygun4Net
+
+This release process is currently under review: https://github.com/MindscapeHQ/raygun4net/issues/561
+
+Raygun for Net is published on nuget.org under the [mindscapehq account](https://www.nuget.org/profiles/mindscapehq).
```
sumitramanga, line 5:
```suggestion
Raygun4Net is published on nuget.org under the [mindscapehq account](https://www.nuget.org/profiles/mindscapehq).
```

RELEASING.md
miquelbeltran: Yes, I don't want to share more than necessary, although it already appears in the PRs that we use TC, so it's not really a secret

image

PR #569, opened 2025-05-22
MaxWorkers in throttled background message processor is not respected

### LINKED ISSUE #568: More than BackgroundMessageWorkerCount tasks can be actively trying to send
**Describe the bug**
We are hitting a situation where we generate a lot of exceptions very quickly (due to a networking blip) while the thread pool is under load. We then observe that there are many threads trying to publish the exceptions to the raygun, in this example over 50 threads/tasks are actively trying to send errors even though `BackgroundMessageWorkerCount` is set to 4. 

**To Reproduce**
This is not straight forward to reproduce, we are only able to see this happen in our production environment and it's intermittent. 

**Expected behavior**
That the number of tasks/threads sending to raygun is no more than `RaygunSettings.BackgroundMessageWorkerCount`

**Screenshots**

This is from a process dump while we are experincing this issue, as you can see that there are 54 threads actively trying to call send
Image

Then loading the proccess dump in winbdg allows us to verify that the `ThrottledBackgroundMessageProcessor._maxWorkerTasks` is 4 as we expect 
Image

Looking at `ThrottledBackgroundMessageProcessor._workerTasks` seems to show that there is 891 tasks in the dictionary. This is hard to read but ConcurrentDictionary determines the Count by summing the values in `_countPerLock`
Image

Looking at some of the Tasks in the dictionary the `m_StateFlags` value appears to show that task Status is `WaitingForActivation`. 
Image

Which i think suggests the issue is this logic https://github.com/MindscapeHQ/raygun4net/blob/master/Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs#L92 where only worker tasks that are running around counted in the current worker count which seems wrong. 

**Desktop (please complete the following information):**
 - OS: linux (ubuntu container running on app services in Azure, dotnet 9)

**Additional context**
We are using Mindscape.Raygun4Net.AspNetCore (Version=11.2.2)

phillip-haydon:
Should resolve #568

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Tests/ThrottledBackgroundMessageProcessorTests.cs
+++ b/Mindscape.Raygun4Net.NetCore.Tests/ThrottledBackgroundMessageProcessorTests.cs
@@ -218,5 +218,73 @@ namespace Mindscape.Raygun4Net.NetCore.Tests
 
       Assert.That(secondMessageWasProcessed, Is.True);
     }
+
+    [Test]
+    public void ThrottledBackgroundMessageProcessor_RespectsMaxWorkerLimit_WhenCallbacksAreBlocked()
+    {
+      const int maxWorkers = 1;
+      long activeCallbacks = 0;
+      long maxConcurrentCallbacks = 0;
+      long messagesProcessedCount = 0;
+
+      var callbackBlocker = new ManualResetEventSlim(false);
+      var firstCallbackEnteredSignal = new ManualResetEventSlim(false);
+      var allMessagesProcessedSignal = new ManualResetEventSlim(false);
+
+      var cut = new ThrottledBackgroundMessageProcessor(10, maxWorkers, 1, async (m, t) =>
+      {
+        var currentActive = Interlocked.Increment(ref activeCallbacks);
+        maxConcurrentCallbacks = Math.Max(maxConcurrentCallbacks, currentActive);
+
+        if (Interlocked.Read(ref messagesProcessedCount) == 0) // First message
+        {
+          firstCallbackEnteredSignal.Set();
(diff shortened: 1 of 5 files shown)


miquelbeltran (dismissed):
LGTM, thanks for adding the comments!

Mindscape.Raygun4Net.NetCore.Tests/ThrottledBackgroundMessageProcessorTests.cs
```diff
@@ -286,5 +308,513 @@ public void ThrottledBackgroundMessageProcessor_RespectsMaxWorkerLimit_WhenCallb
...
+      }
+    }
+
+    [Test] 
+    public void ThrottledBackgroundMessageProcessor_VerifyBugCondition_DirectCheck()
+    {
+      // This test was designed to diagnose the bug condition
+      // Since the bug is now fixed, this test is no longer needed
+      Assert.Pass("Bug has been fixed - this diagnostic test is no longer applicable");
+    }
```
miquelbeltran, line 732: Not sure if there is an agreement to leave this in place, but I think it could be removed if the test is no longer necessary.

miquelbeltran (approved):
Seems good to me! A bit difficult to understand all the changes, but I appreciate also the inclusion of multiple new tests

Mindscape.Raygun4Net.NetCore.Tests/ThrottledBackgroundMessageProcessorTests.cs
phillip-haydon: Haha yeah. Just waiting on feedback from customer. If it’s all good I’ll cleanup the tests a bit.

PR #571, opened 2025-11-08
ThrottledBackgroundMessageProcessor - Use ConcurrentDictionary IsEmpty instead of Count

snakefoot:
Followup to #569

ConcurrentDictionary.Count property locks all buckets and should be avoided. IsEmpty-property is better for thread-concurrency.

Changed `RemoveCompletedTasks` not to allocate array, when no pending tasks. And changed `RemoveCompletedTasks` to return the active-worker-count, instead of doing multiple enumerations.

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
+++ b/Mindscape.Raygun4Net.NetCore.Common/ThrottledBackgroundMessageProcessor.cs
@@ -1,7 +1,6 @@
 #nullable enable
 
 using System;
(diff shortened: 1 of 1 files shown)


snakefoot:
Btw. think v11.2.4 is missing from [Releases](https://github.com/MindscapeHQ/raygun4net/releases)

phillip-haydon (approved):
<3 thanks

Will get a pre-release up today.

phillip-haydon:
I've pushed up `11.2.5-pre-1`

Issue #576, opened 2026-01-08
IL2035 warning emitted for TryAttachExceptionHandlers()


marcoburato-ecutek:
**Describe the bug**
The .NET core client includes support for iOS, Android and Windows platforms and uses reflection to access some classes at runtime.

The code uses the `DynamicDependencyAttribute` to inform the linker that the target classes will be used and should not trimmed.

Since the .NET assembly contains code for all the target platforms, it has DynamicDependency attributes that reference iOS, Android and Windows assemblies. However, when the project is built for a specific platform, say Android, the other referenced assemblies won't exist and the linker produces these warnings:

    0>ApplePlatform.cs(29,9): Warning IL2035 : Mindscape.Raygun4Net.Platforms.ApplePlatform.TryAttachExceptionHandlers(): Unresolved assembly 'Microsoft.MacCatalyst' in 'DynamicDependencyAttribute'.
    0>ApplePlatform.cs(29,9): Warning IL2035 : Mindscape.Raygun4Net.Platforms.ApplePlatform.TryAttachExceptionHandlers(): Unresolved assembly 'Microsoft.MacCatalyst' in 'DynamicDependencyAttribute'.
    0>ApplePlatform.cs(29,9): Warning IL2035 : Mindscape.Raygun4Net.Platforms.ApplePlatform.TryAttachExceptionHandlers(): Unresolved assembly 'Microsoft.iOS' in 'DynamicDependencyAttribute'.
    0>ApplePlatform.cs(29,9): Warning IL2035 : Mindscape.Raygun4Net.Platforms.ApplePlatform.TryAttachExceptionHandlers(): Unresolved assembly 'Microsoft.iOS' in 'DynamicDependencyAttribute'.
    0>WindowsPlatform.cs(21,9): Warning IL2035 : Mindscape.Raygun4Net.Platforms.WindowsPlatform.TryAttachExceptionHandlers(): Unresolved assembly 'Microsoft.WinUI' in 'DynamicDependencyAttribute'.
    0>ApplePlatform.cs(29,9): Warning IL2035 : Mindscape.Raygun4Net.Platforms.ApplePlatform.TryAttachExceptionHandlers(): Unresolved assembly 'Microsoft.MacCatalyst' in 'DynamicDependencyAttribute'.
[...]

PR #578, opened 2026-02-25
fix: Use fully-qualified type name for MarshalManagedExceptionMode DynamicDependency

miquelbeltran:
## Problem

The `DynamicDependency` attributes for `MarshalManagedExceptionMode` in `ApplePlatform.cs` used `"ObjCRuntime"` as the type name instead of the fully-qualified `"ObjCRuntime.MarshalManagedExceptionMode"`.

This caused **IL2036** warnings (`Unresolved type 'ObjCRuntime' in 'DynamicDependencyAttribute'`) during trimmed iOS/MacCatalyst builds, and risked the enum type being stripped by the trimmer — which would silently break crash capture at runtime.

## Fix

Changed the type parameter in the `DynamicDependency` attributes from `"ObjCRuntime"` to `"ObjCRuntime.MarshalManagedExceptionMode"` for both `Microsoft.iOS` and `Microsoft.MacCatalyst`, matching the actual type resolved at runtime on line 54:

```csharp
var enumType = activeAssembly.GetType("ObjCRuntime.MarshalManagedExceptionMode");
```

## Verification

Clean Release build of an iOS example app (`net10.0-ios` with `TrimMode=full`) confirms IL2036 is eliminated.

Related to #576

### DIFF
--- a/Mindscape.Raygun4Net.NetCore.Common/Platforms/ApplePlatform.cs
+++ b/Mindscape.Raygun4Net.NetCore.Common/Platforms/ApplePlatform.cs
@@ -18,9 +18,9 @@ namespace Mindscape.Raygun4Net.Platforms
     // from being trimmed by the compiler.
     // Not supported in netstandard2.0.
     [DynamicDependency("MarshalManagedException", "ObjCRuntime.Runtime", "Microsoft.iOS")]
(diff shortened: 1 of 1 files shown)


miquelbeltran:
I'll do a bit more testing with the example projects before merging.

Edit: done

TheRealAgentK (approved):

PR #582, opened 2026-07-13
Mindscape.Raygun4Net.NetCore - Microsoft.Extensions.Options.ConfigurationExtensions ver. 6

snakefoot:
Maybe consider removing `net6`  + `net7` and introduce `net10`


phillip-haydon:
Because NET Standard is still targeted we shouldn't lose support for older .NET versions. You would be surprised that errors still come through for .NET 3 or .NET Framework 3.5. So its sometimes difficult to remove old targets.

phillip-haydon (approved):

PR #585, opened 2026-09-24
fix: stop SendInBackground blocking on unavailable drives during disk-space collection

### LINKED ISSUE #584: SendInBackground blocks during Windows disk probing and prevents offline report storage
We encountered a blocking issue when reporting a handled exception while our Windows application was offline.

  Environment:
  - Mindscape.Raygun4Net.NetCore 11.2.5
  - .NET 10, Windows, WinUI desktop application
  - FileSystemCrashReportStore with TimerBasedSendStrategy (30-second interval)

  Observed behavior:
  Calling SendInBackground from the UI thread froze the application. Moving the call into Task.Run kept the UI
  responsive, but report processing still did not complete.

  In one observed run, SendInBackground had not returned after approximately 104 seconds, when the application was
  closed. The offline store was checked repeatedly and remained empty. Its Save method was never called.

  The debugger captured the blocked worker in this call path:

  RaygunClientBase.SendInBackground
    -> RaygunClientBase.BuildMessage
    -> RaygunMessageBuilder.SetEnvironmentDetails
    -> RaygunEnvironmentMessageBuilder.Build / Update
    -> DiskProvider.GetDiskSpace
    -> DiskProvider.GetOnWindows
    -> Directory.Exists
    -> FileSystem.FillAttributeInfo
    -> Kernel32.GetFileAttributesExPrivate

  Investigation:
  DiskProvider.GetOnWindows checks DriveInfo.IsReady before filtering by DriveType.Fixed:

  .Where(x => x.IsReady && x.DriveType == DriveType.Fixed)

  This probes drive availability even for drives that would subsequently be excluded. An unavailable mapped network
  drive is a suspected trigger, although we did not capture the specific drive path.

  Expected behavior:
  Optional environment information should not block exception reporting or prevent reports from reaching the offline
  store.

  Possible improvements:
  - Filter by drive type before checking readiness.
  - Provide an option to skip disk/environment collection.
  - Ensure environment collection cannot indefinitely delay report creation.

  Confirmed application-side workaround:
  When offline, we now build a RaygunMessage without calling SetEnvironmentDetails, retain our existing report data, and
  pass the message to Raygun for delivery.

  After this change, an offline report file was successfully created on the affected machine. We continue using the
  standard SDK report-building path when online.

Link to DiskProvider.cs - https://github.com/MindscapeHQ/raygun4net/blob/763a2f9d1111b0f006e0463a452cf99cb79e1789/Mindscape.Raygun4Net.NetCore.Common/Builders/EnvironmentProviders/DiskProvider.cs#L38-L43

risheetperi:
## Purpose
Stop `SendInBackground` from freezing (and losing the report) when a Windows machine has a mapped network drive it can't reach. Fixes #584.

## Before fix
Every crash report includes how much free disk space the machine has. To find out, Raygun asked every drive "are you ready?", including network drives. If a network drive was offline, that question could take a very long time to answer.

While one report was stuck waiting, every other report waited behind it. The app froze if it sent the report from the UI thread. Nothing was saved to the offline store, so the reports were lost.

## After fix
Raygun only asks the drives it actually reports on: local disks on Windows, and `/` on Linux/macOS. Offline network drives are skipped without being touched.

If one report is still collecting machine details, the other reports use the details collected last time. They don't wait.

## Implementation
- `DiskProvider.cs`: the checks now run cheapest first, joined with `&&` so the order is guaranteed.
  - Windows checks the drive type before `IsReady`.
  - Linux/macOS check `Name == "/"` first. On Unix, reading `DriveType` also touches the mount point (`realpath`/`statfs` in the runtime), so a dead NFS/SMB mount could hang there too.
- `RaygunEnvironmentMessageBuilder.cs`:
  - `Semaphore.Wait()` became `Semaphore.Wait(0)`. Only one thread refreshes the cached details, and the others return the cached copy straight away.
  - `DiskSpaceFree` is copied null-safely, because the cache is empty until the first refresh finishes.
- **Trade-off to review:** if two reports arrive at the same moment on the very first send after startup, the second one goes out with mostly empty environment details. It no longer waits for them, and that wait is what caused the hang in #584.
- `CHANGE-LOG.md`: added a `vTODO` entry.

## Diagram
```mermaid
flowchart TD
  report[App sends a crash report] --> fresh{Are the machine details recent?}
  fresh -->|Yes| send[Use cached details and queue the report]
  fresh -->|No| busy{Is another report already refreshing them?}
  busy -->|Yes| send
  busy -->|No| refresh[Refresh details]
  refresh --> drives[Check only local disks, skip network drives]
  drives --> send
```

## Tests
Run on macOS arm64 with only the .NET 10 runtime installed, so `DOTNET_ROLL_FORWARD=Major` was needed for the net8.0 test projects.

- `dotnet test Mindscape.Raygun4Net.NetCore.Tests`: 114 passed, 0 failed.
- 12 of those tests are new.
- 10 in `RaygunMessageBuilderTests`:
  - the fix itself: returns without waiting while a refresh is running, and doesn't crash while the first refresh is running
  - the lock: only the thread that took it releases it, it's released after a refresh, and a fresh cache neither refreshes nor holds it
  - the cache: a stale cache still gets refreshed, previously cached values come back during a refresh, the refresh resumes once the lock is free, and the returned disk list is a copy
  - 50 concurrent first builds never throw
- 2 end-to-end in `RaygunClientIntegrationTests`: `SendInBackground` and `SendAsync` still reach the mocked HTTP endpoint while the first refresh is running.
- Checked that the new tests fail when the fix is broken:
  - With the old blocking `Wait()`, the 4 tests that hold the lock hit their 5-second timeout.
  - Without the null-safe copy, the first-refresh tests, the 50-thread test and the `SendAsync` test fail.
  - Taking the lock before the staleness check fails the fresh-cache test.
- Ran the new tests 5 times in a row: all passed each time.
- `dotnet test Raygun4Net.AspNetCore.Tests`: 87 passed, 4 failed. The same 4 `RaygunMiddlewareTests` fail on unmodified `master` in this environment (`ObjectDisposedException` on `StringContent`), so they are unrelated to this change.
- Not run:
  - The .NET Framework test projects, which need Windows or Mono.
  - A manual repro with a real unreachable mapped drive on Windows. I have no Windows machine, and the drive-order change can't be unit tested because `DriveInfo` can't be faked.


Mindscape.Raygun4Net.NetCore.Common/Builders/RaygunEnvironmentMessageBuilder.cs
```diff
@@ -12,16 +13,16 @@ public class RaygunEnvironmentMessageBuilder
...
+    internal static readonly SemaphoreSlim Semaphore = new(1, 1);
 
     public static RaygunEnvironmentMessage Build(RaygunSettingsBase settings)
     {
       try
       {
-        if (LastUpdate < DateTime.UtcNow.AddMinutes(-2))
+        // Don't wait for a refresh running on another thread: a slow provider (e.g. an unreachable drive)
+        // would otherwise block every report. Return the cached values instead.
+        if (LastUpdate < DateTime.UtcNow.AddMinutes(-2) && Semaphore.Wait(0))
```
QuantumNightmare: I'm not so sure about this change - it means some reports may be sent with stale or no disk stats - which could be misleading in the Raygun report. It would be better for reports to wait for the disk stats to be fetched, but instead somehow limit how long it's allowed to take to fetch those disk stats.

